PR #4, opened 2019-02-14
Make the code build

sbose78:
```
$ dep ensure -v
$ go build -v -o ./bin/devopsconsole-operator cmd/manager/main.go
```


deploy/crds/devconsole_v1alpha1_envdeployment_crd.yaml
```diff
@@ -1,9 +1,9 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  name: envdeployments.devconsole.openshift.io
+  name: envdeployments.devopsconsole-operator.openshift.io
 spec:
-  group: devconsole.openshift.io
+  group: devopsconsole-operator.openshift.io
```
alexeykazakov: I think we should use `devopsconsole.openshift.io` group for all our DevOps Console APIs. Not 'devopsconsole-operator.openshift.io'.

sbose78: Ah right! https://github.com/redhat-developer/devopsconsole-operator/pull/4/commits/d65f2374124b477a156a669c966561b240929bc3

alexeykazakov: Looks good. What about the other places too? `devopsconsole-operator.openshift.io` group is all over the place :)

sbose78: Right https://github.com/redhat-developer/devopsconsole-operator/pull/4/commits/cf023c28b7b7cb31d4302a7e5c7f02da151ae9c0

PR #7, opened 2019-02-22
Add GitSource CRD

alexeykazakov:
See https://jira.coreos.com/browse/ODC-65


kwk:
Can you please leave a comment in the description about what this is meant for or maybe a reference to the JIRA ticket?

deploy/crds/devopsconsole_v1alpha1_gitsource_crd.yaml
```diff
@@ -0,0 +1,41 @@
...
+        apiVersion:
+          type: string
+        kind:
+          type: string
+        metadata:
+          type: object
+        spec:
+          properties:
+            url:
+              type: string
```
kwk: Today I was talking to @tinakurian who is writing the build tool detector. While talking we realized that the Import and Create flow are literally the same thing in the beginning of the flow and end potentially differently. Both flows clone a repo from github in the beginning. In the end, they push either a devfile (import flow) or a complete code package to a git remote.

Now, because of this I suggest to reflect the similarity between these two here.

I cannot recall directly but do you want to use two GitSource CRs for cloning and pushing? If not and you only want to use one, I suggest to have two URLs here, one for pull and one for push. The reason is that for an import flow, we should already know what credentials and transport protocol we need in order to clone a repo. Remember, that during an import, we just clone a github booster repo which will eventually become the new project that we push somewhere else.

Make sense?

@sbose78 @michaelkleinhenz thoughts?

deploy/crds/devopsconsole_v1alpha1_gitsource_crd.yaml
```diff
@@ -0,0 +1,41 @@
...
+            ref:
+              type: string
+            contextDir:
+              type: string
```
kwk: I hope this is meant to be the directory which we intend to use as the source directory. Consider a repo with frontend and backend all in one. If  the GitSource just wants to build the backend, I guess the `contextDir` will be set to `backend`, correct?

Do you want the `contextDir` to be respected by the build tool dectector so that it only looks inside of it and not in e.g. the `frontend` directory?

examples/devopsconsole_v1alpha1_gitsource_cr.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: devopsconsole.openshift.io/v1alpha1
+kind: GitSource
+metadata:
+  name: example-gitsource
+spec:
+  # Add fields here
```
kwk: What about the rest of the fields?

pkg/apis/devopsconsole-operator/v1alpha1/gitsource_types.go
```diff
@@ -0,0 +1,70 @@
...
+	// NoProxy can be used to specify domains for which no proxying should be performed. Optional.
+	NoProxy string `json:"noProxy,omitempty"`
+
+	// SourceSecret is the name of the secret that contains credentials to access the git repo. Optional.
+	SourceSecret string `json:"sourceSecret,omitempty"`
+}
+
+// GitSourceStatus defines the observed state of GitSource
+// +k8s:openapi-gen=true
+type GitSourceStatus struct {
```
kwk: Does it make sense to maybe store the last fetched/seen commit hash? I mean, if the `ref` is `master` it might still be interesting to store the commit hash that `master` resolved to the last time this repo was seen.

deploy/crds/devopsconsole_v1alpha1_gitsource_crd.yaml
alexeykazakov: Replayed here - https://jira.coreos.com/browse/ODC-65?focusedCommentId=70213&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-70213
TL;DR I think we should use multiple GitSource CRs if/when needed.

deploy/crds/devopsconsole_v1alpha1_gitsource_crd.yaml
alexeykazakov: yes, it's just a sub folder. Could be especially useful for monorepos.

pkg/apis/devopsconsole-operator/v1alpha1/gitsource_types.go
alexeykazakov: Do you have any examples of how we could use such info?

examples/devopsconsole_v1alpha1_gitsource_cr.yaml
alexeykazakov: Added other fields to the example - https://github.com/redhat-developer/devopsconsole-operator/pull/7/commits/3c0276e5b85d276681d5829d7fbdee0292856b82

deploy/crds/devopsconsole_v1alpha1_gitsource_crd.yaml
kwk: oh, okay then.

deploy/crds/devopsconsole_v1alpha1_gitsource_crd.yaml
kwk: Then we don't need that information here. We should just focus on the git source and let the build tool decide which folder to use. Or at least it should be in the domain of the build tool and not of the source tool IMHO.

pkg/apis/devopsconsole-operator/v1alpha1/gitsource_types.go
kwk: When you have a git source configured and you wonder why your build still doesn't contain a fix it might be useful to know of the source is at the correct hash.

kwk:
According to https://jira.coreos.com/browse/ODC-65?focusedCommentId=70213&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-70213 we still need the service classification, right? (github, github enterprise, bitbucket, gitlab, etc.)

kwk:
> According to https://jira.coreos.com/browse/ODC-65?focusedCommentId=70213&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-70213 we still need the service classification, right? (github, github enterprise, bitbucket, gitlab, etc.)

Nevermind, I just saw that you've added the `flavor` field in 3c0276e.

examples/devopsconsole_v1alpha1_gitsource_cr.yaml
kwk: Great. Thank you.

deploy/crds/devopsconsole_v1alpha1_gitsource_crd.yaml
alexeykazakov: What if there are multiple apps in the same repo and they should be build independently? A monorepo. Basically speaking why if some git repo contains multiple independent git sources?
IMO we should let users to specify that optional subfolder.

alexeykazakov: What if there are multiple apps in the same repo and they should be build independently? A monorepo. Basically speaking why if some git repo contains multiple independent git sources?
IMO we should let users to specify that optional subfolder.

pkg/apis/devopsconsole-operator/v1alpha1/gitsource_types.go
alexeykazakov: Ok I see your point but I’m not sure if we should keep this info here or somewhere else. Let’s introduce it later when we need it.

deploy/crds/devopsconsole_v1alpha1_gitsource_crd.yaml
kwk: @alexeykazakov if there are multiple apps in a mono-repo that you want to build independently I'd say we should be able to re-use a `GitSource` in two build configurations and have each build configuaration define a dfifferent `contextDir`. @sbose78 thoughts?

pkg/apis/devopsconsole-operator/v1alpha1/gitsource_types.go
alexeykazakov: I'm not arguing with that possible use case. I just don't think we really have to introduce this now. It's not going to be used anytime soon. I'm not sure if it's really the best place to be. So, it's premature for now. We can considering adding this field later when we are ready to support tracking commits.

deploy/crds/devopsconsole_v1alpha1_gitsource_crd.yaml
sbose78: @kwk 
The buildConfig object would be more of an 'implementation detail' which the controller creates on finding that a new gitSource object has been created ( either as `oc create gitsource` or from the UI ).

So the initial information needs to be captured in the initial object ( gitSource CR ) that gets created.

pkg/apis/devopsconsole-operator/v1alpha1/gitsource_types.go
sbose78: Having the hash stored somewhere looks very interesting.
Hi @kwk, Is this the kind of flow you have in mind?

- A gitSource CR is created and a component CR is created ( which references the gitSource CR ).
- The controller listening on the creation of a new  component CR, creates the buildConfig ( and with it a lot of other things ).
- The 'watch' loop gets to know that a buildConfig *belonging* to the component was created. That could write some status regarding the buildConfig to the `component` CR.
- When a new `build` is created, there's a block which says:
```
revision:
    git:
      author:
        email: sbose78@gmail.com
        name: Shoubhik Bose
      commit: a4743fd8644577b4a0457a552825cb4f1bdd7f3a
      committer:
        email: noreply@github.com
        name: GitHub
      message: parameterize db name
    type: Git
```

- So, if we 'listen' on "Builds in namespace X, not necessarily owned by component C", then we could get this commit information and add it to the `component` status.

Is that what you are saying?

deploy/crds/devopsconsole_v1alpha1_gitsource_crd.yaml
kwk: Okay, after some talking to @michaelkleinhenz I think I get that GitSource isn't meant to represent a cloned repo. It is just the information on how to get to the source. So if there's a frontend and a backend that both build from the same repo, they should have different GitSource objects that point to the same git. But since it is the build teams job to pick a directory that shall be built, it should be their responsibility to configure the contextDir. It is not part of a representation of how to access a git source which is what GitSource is doing in the end.

alexeykazakov: > It is not part of a representation of how to access a git source which is what GitSource is doing in the end.

I'm not sure I'm following... 
Let's say I have a repo with a hundred of apps there. Each of them are supposed to built independently. Don't ask me why I decided to organize my repo in that way. I just like it.
Now I want to import my `<githost>/coolapp` golang app. Edit it in Che. Build it and deploy. I do not want to import my other nodejs app `<githost>/coolapp-doc` which is just a legacy documentation app. How build/che etc are supposed to guess what I really want to import/edit/build without me, as a user, explicitly point to `<githost>/coolapp` when importing the app?

deploy/crds/devopsconsole_v1alpha1_gitsource_crd.yaml
```diff
@@ -0,0 +1,43 @@
...
+            httpProxy:
+              type: string
+            httpsProxy:
+              type: string
+            noProxy:
+              type: string
+            sourceSecret:
+              type: string
```
MatousJobanek: I would probably store secrets separated and have `sourceSecret` as a reference to this CR (eg. `GitSecret` or in general `SourceSecret`). The reason is that the secret is usually associated more with a user account than with a repo. If we had `GitSecret` separated it would make changes/updates of the secret much easier across all `GitSources`. 
Imagine that one user has multiple `GitSource` CRs for several repositories/branches. This user has one GH token that uses for all of the `GitSources` and after some time he would need to change the token. To be honest, I'm not sure how we would manage such a change across all `GitSources` if we didn't have the secret stored seperated.

sbose78: Yes, this should have the name of the `sourceSecret`. Are you suggesting we should have the `type` as `kind` `secret` ?

MatousJobanek: Either secret or more likely a new custom resource. For example something like this:
```yaml
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
...
spec:
  group: devopsconsole.openshift.io
  names:
    kind: SourceSecret
    ...
  validation:
    openAPIV3Schema:
      properties:
        apiVersion:
          type: string
        kind:
          type: string
        metadata:
          type: object
        spec:
          properties:
            flavor:
              type: string
            secret:
              type: secret
```

alexeykazakov: `sourceSecret` is not the credentials itself. It's just a name of the secret in the same namespace. So, basically it's already a reference to a secret.
Maybe `sourceSecret` is not the best name for this field because it can confuse. Will change it to something more clear.
The secret itself will keep different type of Git credentials. SSH key, token (for github for example), certificate, etc. Currently OpenShift build already can handle it and use the secret properly and I think we should leverage it.

kwk: What about `secretRef` as the name?

alexeykazakov: Yeah.. `secretRef` looks good to me.

alexeykazakov: Changed it to
```
  secretRef:
    name: mysecret
```

MatousJobanek: Ok, I misundestood that based of the name and its original value (only string). Thanks for changing it @alexeykazakov, it looks much better now

ljelinkova:
I've noticed that you use `group: devopsconsole.openshift.io`. Is `openshift.io` something that we want to reference for example in 5 years? Wouldn't something with `redhat-developer` be a better fit?

alexeykazakov:
openshift.io is currently used by all OpenShift resources. It doesn’t refers to OSIO. It actually refers to OpenShift Platform.
And DevOps console will be part of OpenShift.

ljelinkova:
Thank you for the clarification.

kwk:
> openshift.io is currently used by all OpenShift resources. It doesn’t refers to OSIO. It actually refers to OpenShift Platform.
> And DevOps console will be part of OpenShift.

@alexeykazakov but still shouldn't we be using something else? I mean the name openshift.io is really taken and what we provide here has literally nothing to do with it. We have a bag of domains to choose from, don't we (e.g. codeready, openshift).

MatousJobanek:
> but still shouldn't we be using something else? I mean the name openshift.io is really taken and what we provide here has literally nothing to do with it. We have a bag of domains to choose from, don't we (e.g. codeready, openshift).

@kwk IMHO, if we want to contribute to openshift (which DevOps console theoretically is the case) then we should be consistent with those names they use in their resources. Look at https://github.com/openshift/origin/tree/master/api/docs - they use `openshift.io` suffix no matter what the actual `openshift.io` domain is used for. Btw, the `.io` suffix is originally taken from k8s

PR #12, opened 2019-03-04
Remove environment related CRDs

sbose78:
Note, these should be brought back ( if possible as a `revert` ) when we add functionality around these CRDs.


sbose78:
Thank you @alexeykazakov ! Had missed them :)

PR #15, opened 2019-03-06
CI job for devops console operator

tinakurian:
Created a CI job for devops console operator which will perform the following steps: 

1. docker-check-go-format
2. docker-deps
3. docker-check-go-code
4. docker-build
5. docker-test-unit

Once this is done a tag will be added to the docker image and push will be attempted to quay (see function deploy() and function tag_push() in cico_setup.sh).

The Makefile was updated in order to have the .cache directory created in a location which permissions are available for. Without specifying the directory by setting  XDG_CACHE_HOME the .cache directory will fail to be created in the docker container as it will attempt to do so in a location which requires root permissions. This will cause the docker-check-go-code (step 3) to fail if not set appropriately.


sbose78:
This PR has the CD script. The CI script is missing.
Add the cico_run_tests.sh which would just compile the code and run tests, but not push to quay?

tinakurian:
> This PR has the CD script. The CI script is missing.
> Add the cico_run_tests.sh which would just compile the code and run tests, but not push to quay?

Got it! Done here 1df8e7d36b011dfffcef3de7c4f0747022e9eb9a

cico_setup.sh
```diff
@@ -0,0 +1,105 @@
...
+function run_tests_with_coverage() {
+  # Run the unit tests that generate coverage information
+  make docker-test-unit-with-coverage
+
+  # Output coverage
+  make docker-coverage-all
+
+  # Upload coverage to codecov.io
+  cp tmp/coverage.mode* coverage.txt
+  bash <(curl -s https://codecov.io/bash) -X search -f coverage.txt -t 6539a1b6-70f9-4ba1-8ab5-d1c8a104ec83 #-X fix
```
kwk: @tinakurian when you continue to use this token you will be pushing coverage information to the toolchain-operator project because that uses the same hash (https://github.com/fabric8-services/toolchain-operator/blob/83f2e947ad81bcdf30beb507c5cb589ac9c74ded/cico_setup.sh#L62).

tinakurian: Thanks Konrad, addressed here: https://github.com/redhat-developer/devopsconsole-operator/pull/17

PR #20, opened 2019-03-11
Component controller should watch all namespaces

jarifibrahim:
This PR changes the `manager.Options` value to set `namespace` to `''`.
With this change, the component operator will watch all namespaces.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -10,7 +10,6 @@ import (
 	appsv1 "github.com/openshift/api/apps/v1"
 	buildv1 "github.com/openshift/api/build/v1"
 	imagev1 "github.com/openshift/api/image/v1"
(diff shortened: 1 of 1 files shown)


corinnekrych:
@jarifibrahim you need to remove:
https://github.com/redhat-developer/devopsconsole-operator/blob/2bc2993b80172a2bef126e981dec151bc49d2e53/cmd/manager/main.go#L45-L49
to make the code compile.
But perfect!
it was exactly what i needed to be able to see the image stream in `openshift` namespace 🎉

jarifibrahim:
@corinnekrych Done in https://github.com/redhat-developer/devopsconsole-operator/pull/20/commits/f4c924c19861550f2c1b5b396f44d6bfcd55cd39

PR #21, opened 2019-03-11
Use builder image from the openshift namespace

corinnekrych:
fixes https://jira.coreos.com/browse/ODC-133
if runtime image is found on openshift namespace use it, instead of pulling from docker hub.
For unit-test i took ex on Dipak work on toolchain-operator

### DIFF
--- a/pkg/apis/addtoscheme_devopsconsole_v1alpha1.go
+++ b/pkg/apis/addtoscheme_devopsconsole_v1alpha1.go
@@ -1,7 +1,7 @@
 package apis
 
 import (
-	"github.com/redhat-developer/devopsconsole-operator/pkg/apis/devopsconsole/v1alpha1"
+	"github.com/redhat-developer/devopsconsole-operator/pkg/apis/devopsconsole-operator/v1alpha1"
 )
 
 func init() {
--- a/pkg/controller/component/component_controller.go
+++ b/pkg/controller/component/component_controller.go
@@ -163,7 +163,7 @@ func (r *ReconcileComponent) getBuilderImage(instance *componentsv1alpha1.Compon
 	found := &imagev1.ImageStream{}
 	err := r.client.Get(context.TODO(), types.NamespacedName{Name: instance.Spec.BuildType, Namespace: openshiftNamespace}, found)
 	if err != nil {
-		log.Info(fmt.Sprintf("** Searching in namespace %s imagestream %s fails **"), openshiftNamespace, instance.Spec.BuildType)
+		log.Info(fmt.Sprintf("** Searching in namespace %s imagestream %s fails **", openshiftNamespace, instance.Spec.BuildType))
 		// Create an empty image name "nodejs-runtime"
 		newImageForRuntime = newImageStreamFromDocker(instance.Namespace, instance.Name, instance.Spec.BuildType)
 		if newImageForRuntime == nil {
--- a/pkg/controller/component/component_controller.go
+++ b/pkg/controller/component/component_controller.go
@@ -118,7 +118,7 @@ func (r *ReconcileComponent) Reconcile(request reconcile.Request) (reconcile.Res
 	// We only call the pipeline when the component has been created
 	// and if the Status Revision Number is the same
 	if instance.Status.RevNumber == instance.ObjectMeta.ResourceVersion {
-		// Create an empty image name "nodejs-output"
+		// Create an empty image name "myapp-output"
 		newImageForOutput := newImageStream(instance.Namespace, instance.Name)
 		err = r.client.Create(context.TODO(), newImageForOutput)
 		if err != nil {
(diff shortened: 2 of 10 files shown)


pkg/controller/component/component_controller.go
```diff
@@ -127,27 +130,15 @@ func (r *ReconcileComponent) Reconcile(request reconcile.Request) (reconcile.Res
 			log.Error(err, "** Setting owner reference fails **")
 			return reconcile.Result{}, err
 		}
-
-		// Create an empty image name "nodejs-runtime"
-		// todo(corinne): check if we can reuse openshift's image stream
-		newImageForRuntime := newImageStreamFromDocker(instance.Namespace, instance.Name, instance.Spec.BuildType)
-		err = r.client.Create(context.TODO(), newImageForRuntime)
+		// Create a build image named either "myapp-runtime" or reuse openshift "nodejs" runtime
+		ir, err := r.getImageRuntime(instance)
```
sbose78: Could you please call it `getBuilderImage(..)` ? The reason I try to specify the difference between `builder` and `runtime` is that some languages do not have the same image requirements at build and at runtime.

Example, Golang apps need a golang builder image but doesn't need a go runtime.

pkg/controller/component/component_controller_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+		//then
+		require.NoError(t, err, "reconcile is failing")
+
+		instance := &compv1alpha1.Component{}
+		errGet := r.client.Get(context.TODO(), req.NamespacedName, instance)
+		require.NoError(t, errGet, "component is not created")
+
+		is := &imagev1.ImageStream{}
+		errGetImage := cl.Get(context.Background(), types.NamespacedName{Namespace: Namespace, Name: Name + "-output"}, is)
```
sbose78: I might be missing something, are you validating the content of `is` ,`bc`..etc ?

pkg/controller/component/component_controller_test.go
```diff
@@ -0,0 +1,185 @@
...
+		require.NoError(t, errGetImage, "output imagestream is not created")
+
+		isRuntime := &imagev1.ImageStream{}
+		errGetRuntimeImage := cl.Get(context.Background(), types.NamespacedName{Namespace: Namespace, Name: Name + "-runtime"}, isRuntime)
+		require.NoError(t, errGetRuntimeImage, "runtime imagestream is not created")
+
+		bc := &buildv1.BuildConfig{}
```
sbose78: Could you validate the content of `bc` ?

pkg/controller/component/component_controller.go
corinnekrych: fixed in 6a1332a

corinnekrych: 6a1332a

sbose78:
[test]

sbose78:
[test]

pkg/controller/component/component_controller_test.go
corinnekrych: some validation are done a few line after but let me add more

corinnekrych: 73312f8

sbose78:
While reviewing the code, I found a lot of log messages with `nodejs` hardcoded. Please fix that.

pkg/controller/component/component_controller.go
```diff
@@ -127,27 +130,15 @@ func (r *ReconcileComponent) Reconcile(request reconcile.Request) (reconcile.Res
 			log.Error(err, "** Setting owner reference fails **")
 			return reconcile.Result{}, err
 		}
-
-		// Create an empty image name "nodejs-runtime"
-		// todo(corinne): check if we can reuse openshift's image stream
-		newImageForRuntime := newImageStreamFromDocker(instance.Namespace, instance.Name, instance.Spec.BuildType)
-		err = r.client.Create(context.TODO(), newImageForRuntime)
+		// Create a build image named either "myapp-builder" or reuse openshift "nodejs" builder image
+		ir, err := r.getBuilderImage(instance)
```
sbose78: An improvement to the code style would be to define 'builder image resolvers' and pass them elegantly - such that the first attempt would be to lookup in the `openshift` namespace and then the static list.

pkg/controller/component/component_controller.go
```diff
@@ -56,12 +56,15 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 	return nil
 }
 
-var _ reconcile.Reconciler = &ReconcileComponent{}
-var buildTypeImages = map[string]string{"nodejs":"nodeshift/centos7-s2i-nodejs:10.x"}
+
+var (
+	_                  reconcile.Reconciler = &ReconcileComponent{}
+	buildTypeImages                        = map[string]string{"nodejs": "nodeshift/centos7-s2i-nodejs:10.x"}
+	openshiftNamespace                      = "openshift"
```
sbose78: Let's move this to a different file. I'm expecting this list to grow to at 6-8 images.

pkg/apis/addtoscheme_devopsconsole_v1alpha1.go
```diff
@@ -1,7 +1,7 @@
 package apis
 
 import (
-	"github.com/redhat-developer/devopsconsole-operator/pkg/apis/devopsconsole/v1alpha1"
+	"github.com/redhat-developer/devopsconsole-operator/pkg/apis/devopsconsole-operator/v1alpha1"
```
sbose78: https://chat.openshift.io/developers/pl/zjz4xey3nfgs3yej1rd3ikutiw :)

corinnekrych:
7a4b641 to remove mention of nodejs

pkg/apis/addtoscheme_devopsconsole_v1alpha1.go
corinnekrych: agreed

pkg/controller/component/component_controller.go
corinnekrych: Good idea, we can do it in a separate PR.
first let's add unit tests with this Pr
and then odc-132 to add e2e test (plus some RBAC correction needed)
Once we got a good test coverage, refactor the code for more elegant approach.

sbose78: OK, please create issues and track it. Add a comment in the code too.

corinnekrych: @sbose78 https://jira.coreos.com/browse/ODC-242 crested

PR #22, opened 2019-03-12
configure operator.yaml and local env to watch all namespaces

corinnekrych:
as we've discussed in MM https://chat.openshift.io/developers/pl/ujpneqe4gjda9x16b9gegdx1nh

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -35,7 +35,7 @@ spec:
             - name: WATCH_NAMESPACE
               valueFrom:
                 fieldRef:
-                  fieldPath: metadata.namespace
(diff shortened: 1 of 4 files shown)


deploy/operator.yaml
```diff
@@ -35,7 +35,7 @@ spec:
             - name: WATCH_NAMESPACE
               valueFrom:
```
xcoulon: you can use 
```
- name: WATCH_NAMESPACE
  value: ""
```
which is even simpler

corinnekrych: fixed in aabe21d

PR #28, opened 2019-03-15
make the operator watching all namespaces

corinnekrych:
I need your input, trying the cluster scope for the operator without success :'(
- See the step in README
- the error I get:
```
Creating new OUTPUT image fails **","error":"an empty namespace may not be set during creation","stacktrace":"github.com/redhat-developer/devopsconsole-operator/vendor/github.com/go-logr/zapr.(*zapLogger).Error\n\t/Users/corinne/workspace/go/src/github.com/redhat-developer/devopsconsole-operator/vendor/github.com/go-logr/zapr/zapr.go:128\ngithub.com/redhat-developer/devopsconsole-operator/pkg/controller/component.(*ReconcileComponent).Reconcile\n\t/Users/corinne/workspace/go/src/github.com/redhat-developer/devopsconsole-operator/pkg/controller/component/component_controller.go:125\ngithub.com/redhat-developer/devopsconsole-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/Users/corinne/workspace/go/src/github.com/redhat-developer/devopsconsole-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:207\ngithub.com/redhat-developer/devopsconsole-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1\n\t/Users/corinne/workspace/go/src/github.com/redhat-developer/devopsconsole-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:157\ngithub.com/redhat-developer/devopsconsole-operator/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/Users/corinne/workspace/go/src/github.com/redhat-developer/devopsconsole-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133\ngithub.com/redhat-developer/devopsconsole-operator/vendor/k8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/Users/corinne/workspace/go/src/github.com/redhat-developer/devopsconsole-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:134\ngithub.com/redhat-developer/devopsconsole-operator/vendor/k8s.io/apimachinery/pkg/util/wait.Until\n\t/Users/corinne/workspace/go/src/github.com/redhat-developer/devopsconsole-operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88"}
```
it looks like the CR does not have a namespace associated to it....


corinnekrych:
Following documentation in https://github.com/operator-framework/operator-sdk/blob/master/doc/user-guide.md I've done a small poc repo: https://github.com/corinnekrych/component-operator with the `--cluster-scoped` option when generating the project, then I compare the `role.yaml` and `role_binding.yaml` file and apply exactly the same pattern here.

@tinakurian if you follow the instruction in README, you should be able to deploy the operator and use it in any namespace. I use minishift internal registry, but eventually we should change that and use the image published by CI. we might need to adjust Makefile to cover both cases. this PR focus on rbac and cluster scope with a manual approach to start with.

@xcoulon fun fact if i replace `ClusterRole` by `Role` in `role.yaml` it stills works... but I've stick to what was produce by scaffolding.

deploy/service_account.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: devopsconsole-operator
+  name: devopsconsole-operator
```
baijum: Please change your editor configuration not to remove the newline.
https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline/729795#729795

corinnekrych: Indeed, let me fix that.
@baijum I found that cross-IDE formatting: https://editorconfig.org/ that may help. wdyt?

deploy/operator.yaml
```diff
@@ -16,21 +16,10 @@ spec:
       containers:
         - name: devopsconsole-operator
           # Replace this with the built image name
-          image: REPLACE_IMAGE
-          ports:
-          - containerPort: 60000
-            name: metrics
+          image: 172.30.1.1:5000/devopsconsole/devopsconsole-operator:latest
```
baijum: Is this IP going to be the same everywhere?

corinnekrych: Yes this is minishift internale registry. 
But note this PR is to show how to deploy operator manually so that you can use CR in _any_ namespace.
We'll do another Pr for e2e test focusing on CI/CD deployment and running e2e test. Most probably this README section will be refactored with some easier makefile target.

corinnekrych:
@xcoulon In the end, I think ClusterRole is not needed.
What helped me discovering it is to use: `--cluster-scoped` option to generate the right Role/RoleBinding but then removing the non needed `Cluster` prefix.
I think the way operator-sdk's templates are generated use a confusing naming.

xcoulon:
> @xcoulon In the end, I think ClusterRole is not needed.
> What helped me discovering it is to use: `--cluster-scoped` option to generate the right Role/RoleBinding but then removing the non needed `Cluster` prefix.
> I think the way operator-sdk's templates are generated use a confusing naming.

ok thanks for clarifying this, @corinnekrych

deploy/operator.yaml
dipak-pawar: @corinnekrych I think you don't need to hard code this, you can do `eval $(minishift docker-env)` and then build docker image, and use the same image without pushing it to minishift internal registry, I believe it'll be pain if you are pulling, pushing each time for each change. we are doing the same for toolchain operator.
`

README.md
```diff
@@ -81,6 +80,56 @@ NAME                                  TYPE      FROM          STATUS    STARTED
 build.build.openshift.io/myapp-bc-1   Source    Git@85ac14e   Running   45 seconds ago
 ```
 
+### Deploy the operator with Deployment yaml
+
+* (optional) minishift internal registry
+Build the operator's controller image and make it available in internal registry
+```
+oc adm policy add-cluster-role-to-user cluster-admin admin
```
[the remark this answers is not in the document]
corinnekrych: it's to simplify the docker push (which need a password) 
Overall those instructions will be removed from README in favor of makefile targets

corinnekrych:
@dipak-pawar @xcoulon I've updated to the minimal role needed. Ive tested it with a dichotomy approach.

README.md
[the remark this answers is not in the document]
dipak-pawar: see my  below comments, do we really need docker push?

Makefile
```diff
@@ -311,11 +311,11 @@ deploy-operator: deploy-crd
 .PHONY: deploy-clean
 ## Deploy a CR as test
 deploy-clean:
-	@-oc delete imagestream.image.openshift.io/myapp-runtime
+	@-oc delete component.devopsconsole.openshift.io/myapp
```
dipak-pawar: this won't execute all remaining commands if failed for the first command. maybe appending `|| true` will help here

deploy/role_binding.yaml
```diff
@@ -1,11 +1,12 @@
-kind: RoleBinding
+kind: ClusterRoleBinding
```
dipak-pawar: rename file name to `cluster_role_binding.yaml`

deploy/role.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
+kind: ClusterRole
```
dipak-pawar: maybe worth to remove file name to `cluster_role.yaml`, it looks confusing to me

deploy/test/global-manifests.yaml
```diff
@@ -21,4 +21,4 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: devconsole
-  namespace: devconsole-e2e-test
\ No newline at end of file
+  namespace: devconsole-e2e-test
```
dipak-pawar: I think we need to update this global-manifests.yaml with updated cluster role

README.md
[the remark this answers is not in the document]
corinnekrych: yes otherwise the image is in the cash but not in the internal registry, and you end up with a ImagePull error when deployinh the operator

deploy/test/global-manifests.yaml
corinnekrych: yes but that would be another PR for now i focus on manual deployment
odc-132 is to deal with e2e test

README.md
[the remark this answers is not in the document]
dipak-pawar: if you can set `imagePullPolicy: IfNotPresent` in your pod spec for your local dev/test scenario. and we would be out from this issue. https://kubernetes.io/docs/concepts/containers/images/

deploy/test/global-manifests.yaml
dipak-pawar: sure

Makefile
corinnekrych: @-oc => carry on if failed

deploy/role.yaml
corinnekrych: done in 930cb5a

deploy/operator.yaml
corinnekrych: you need to push to the internal registry.

Makefile
xcoulon: yes, the `-` prefix can be used to ignore the command exit status code: https://www.gnu.org/software/make/manual/make.html#Errors-in-Recipes

deploy/role.yaml
jarifibrahim: @corinnekrych @dipak-pawar 
With this change, we cannot add a new API because operator sdk complains of the missing `role.yaml` file
```
➜  devopsconsole-operator git:(devopsconsole-installer) ✗ operator-sdk add api --api-version=devopsconsole.openshift.io/v1alpha1 --kind=Installer            
INFO[0000] Generating api version devopsconsole.openshift.io/v1alpha1 for kind Installer. 
INFO[0000] Created pkg/apis/devopsconsole/v1alpha1/installer_types.go 
INFO[0000] Created pkg/apis/addtoscheme_devopsconsole_v1alpha1.go 
INFO[0000] Created deploy/crds/devopsconsole_v1alpha1_installer_cr.yaml 
INFO[0001] Created deploy/crds/devopsconsole_v1alpha1_installer_crd.yaml 
Error: failed to update the RBAC manifest for the resource (devopsconsole.openshift.io/v1alpha1, Installer): (failed to read role manifest /home/ijarif/Projects/go/src/github.com/redhat-developer/devopsconsole-operator/deploy/role.yaml: open /home/ijarif/Projects/go/src/github.com/redhat-developer/devopsconsole-operator/deploy/role.yaml: no such file or directory)
Usage:
  operator-sdk add api [flags]

Flags:
      --api-version string   Kubernetes APIVersion that has a format of $GROUP_NAME/$VERSION (e.g app.example.com/v1alpha1)
  -h, --help                 help for api
      --kind string          Kubernetes resource Kind name. (e.g AppService)

```
The operator-sdk looks for `role.yml` file. See https://github.com/operator-framework/operator-sdk/blob/master/commands/operator-sdk/cmd/add/api.go#L119 and https://github.com/operator-framework/operator-sdk/blob/master/pkg/scaffold/role.go#L33

I would like to rename the `cluster_role.yaml` to `role.yaml` and I've raised https://github.com/redhat-developer/devopsconsole-operator/pull/30 for it. @corinnekrych can you please take a look at that PR?

PR #31, opened 2019-03-21
[Hold] Deploy 'backing' App Service REST API from devops-console operator

jarifibrahim:
### Installing 'backing' API services using the Dev Console Operator

This PR deploys a REST Service called "App Service" using the Dev Console operator, to expose complex capabilities to the UI.

Depends on https://github.com/redhat-developer/devopsconsole-operator/pull/30
See story https://jira.coreos.com/browse/ODC-213

#### Deploy the service

This PR adds new CRD `Installer` and an associated controller. It ensures there is a deployment of `app-service` running.

The controller watches
1. The `installer` CR
2. `Deployments`
3. `Services`
4. `Routes`

Once a `Installer` CR is created, there would be a `deployment`, `service` and `route` created for it. 

#### Publish the route in the `status` of the CR
```
status:
  AppServiceURL: appservice-route-olm.192.168.42.240.nip.io
```

#### Ensure client/UI can discover it.
 
Creates a clusterrole and a clusterrolebinding so that any user can list and read the status of the installer CR.

#### Start operator locally with 
```
$ export APP_SERVICE_IMAGE_NAME="jarifibrahim/boilerplate-app"
$ operator-sdk up local
```
This command would start the operator locally
####  Create a new CR using
```
apiVersion: devopsconsole.openshift.io/v1alpha1
kind: Installer
metadata:
  name: main-installer
  namespace: olm
spec: {}
```

- This would create a new deployment of `jarifibrahim/boilerplate-app` image and expose it via a route.
- The `route` would be added to the `installer` CR
The UI would be able to find the URL of App Service from the `status` of the `installer` CR.

## How would the UI read the route?

1. Every user has read/list access to `installer` CR

#### Login as `developer`

```
Using project "myproject".
➜  devopsconsole-operator git:(devopsconsole-installer) oc whoami                         
developer
```

#### Simulate a call from the UI to get the installer CR `status` information to get the generated route URL

```
➜  devopsconsole-operator git:(devopsconsole-installer) oc get installers.devopsconsole.openshift.io  --all-namespaces            
NAMESPACE   NAME             KIND
olm         main-installer   Installer.v1alpha1.devopsconsole.openshift.io
➜  devopsconsole-operator git:(devopsconsole-installer) oc get installers.devopsconsole.openshift.io main-installer -o yaml -n olm
apiVersion: devopsconsole.openshift.io/v1alpha1
kind: Installer
metadata:
  creationTimestamp: 2019-03-27T12:21:31Z
  generation: 1
  name: main-installer
  namespace: olm
  resourceVersion: "101094"
  selfLink: /apis/devopsconsole.openshift.io/v1alpha1/namespaces/olm/installers/main-installer
  uid: da9ea22f-508a-11e9-b323-525400243fb7
spec: {}
status:
  AppServiceURL: appservice-route-olm.192.168.42.240.nip.io
➜  devopsconsole-operator git:(devopsconsole-installer) 

```

```
curl -X GET http://appservice-route-olm.192.168.42.240.nip.io/status\?format\=yaml
build_time: "2019-03-18T10:50:40Z"
commit: 5a551bb4096f6d976552c98a83bd4cca499b083b
start_time: "2019-03-27T12:21:39Z"

```

Todo

- [x] Update PR description
- [x] Automatically create `installer` CR on operator initialization


pkg/controller/installer/installer_controller.go
```diff
@@ -0,0 +1,372 @@
...
+}
+
+func getLabelsForCR(cr *devopsconsolev1alpha1.Installer) map[string]string {
+	labels := map[string]string{
+		"app": cr.Name,
+	}
+	return labels
+}
+
+// newDeploymentForCR returns a busybox pod with the same name/namespace as the cr
```
sbose78: Update the comment , please

deploy/operator.yaml
```diff
@@ -29,3 +29,5 @@ spec:
                   fieldPath: metadata.name
             - name: OPERATOR_NAME
               value: "devopsconsole-operator"
+            - name: APP_SERVICE_IMAGE_NAME
+              value: "jarifibrahim/boilerplate-app" # change this to actual image when we have it
```
jarifibrahim: This has to be changed to actual image when we have it

deploy/crds/devopsconsole_v1alpha1_installer_cr.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: devopsconsole.openshift.io/v1alpha1
+kind: Installer
+metadata:
+  name: main-installer
+spec: {}
```
sbose78: Can we modify the CRD to support listing which service is to be installed?

At the CR, we could specify the name of the service's image's environment variable example "APP_SERVICE" , and the controller could simply use that information to get the image name?

sbose78: The above comment is intended to help add multiple services - and to do so all someone would needs to do is:

- add environment variable to the operator with the image name.
- create a new installer CR referring to the image name by the environment variable.

jarifibrahim: >At the CR, we could specify the name of the service's image's environment variable example "APP_SERVICE" , and the controller could simply use that information to get the image name?

What would happen if the `APP_SERVICE` entry was removed from the CR? Should the deployment, service, routes should also be deleted?

```yaml
spec:
  services:
    - name: "app-service"
    - name: "another-service"
```
Would the new yaml look like this?

deploy/operator.yaml
kwk: Do we have a repo in quay? Then I can create one. As for the redhat-developer/app-service, it is already built for every new master revision on docker hub: https://cloud.docker.com/u/konradkleine/repository/docker/konradkleine/app-service

sbose78: I've opened this PR to create the repos on Quay https://gitlab.cee.redhat.com/service/app-interface/merge_requests/238/diffs#71dd9efa997ab16894984e6781c01dc898af5e13_0_33

deploy/crds/devopsconsole_v1alpha1_installer_cr.yaml
sbose78: > What would happen if the APP_SERVICE entry was removed from the CR? Should the deployment, service, routes should also be deleted?

Yes.


> Would the new yaml look like this?

Yes. Though you need to find a way to find a connection between the "app-service" and the environment variable that holds the image for that.

sbose78:
Any thoughts on  how we would be bundling the extra rolebindings?

sbose78:
```
- rules:
        apiGroups:
        - devopsconsole.openshift.io
        resources:
        - installers
        verbs:
        - get
        systemGroupName: system:authenticated
```

Have to test this.

sbose78:
That wouldn't work :)

https://github.com/operator-framework/operator-lifecycle-manager/blob/9febd60fde1837bf510308bbd6a5f10fed53c7ab/pkg/controller/install/deployment.go#L20

jarifibrahim:
> That wouldn't work :)
> 
> https://github.com/operator-framework/operator-lifecycle-manager/blob/9febd60fde1837bf510308bbd6a5f10fed53c7ab/pkg/controller/install/deployment.go#L20

This should be fixed https://github.com/redhat-developer/devconsole-operator/pull/31/commits/48017c7c65f46cedce232e11ae00e92bf289cfc0#diff-9e63f7b0791edcb4a53c944552b88970R129

deploy/crds/devopsconsole_v1alpha1_installer_cr.yaml
jarifibrahim: We should do this in a separate PR. This PR is already huge.

cmd/manager/main.go
```diff
@@ -103,12 +110,38 @@ func main() {
...
+	if err := createClusterRoleAndBinding(mgr.GetClient()); err != nil {
+		log.Error(err, "")
+		os.Exit(1)
+	}
+	// Create a new "installer". This will trigger the deployment and create route
+	c := mgr.GetClient()
+	installer := &devopsconsolev1alpha1.Installer{}
+	installer.ObjectMeta.Name = "main-installer"
+	installer.ObjectMeta.Namespace = namespace
+	fmt.Println("aaaa", installer.ObjectMeta.Namespace)
```
corinnekrych: to be removed

deploy/role.yaml
```diff
@@ -44,3 +44,22 @@ rules:
...
+  resources:
+  - routes
+  verbs:
+  - "*"
+- apiGroups:
+  - apps
+  resources:
+  - deployments
+  verbs:
+  - "*"
```
corinnekrych: `get` and `create` could be enough. to be tried.

jarifibrahim: Yes. Fixed in https://github.com/redhat-developer/devconsole-operator/pull/31/commits/d83160159c2d7d05b018051d6e22408359e952a6

deploy/operator.yaml
jarifibrahim: > https://gitlab.cee.redhat.com/service/app-interface/merge_requests/238/diffs#71dd9efa997ab16894984e6781c01dc898af5e13_0_33

This PR was closed a few days ago. Are we publishing the image somewhere?

pkg/controller/installer/installer_controller.go
jarifibrahim: Done

deploy/role.yaml
jarifibrahim: Done in https://github.com/redhat-developer/devconsole-operator/pull/31/commits/d83160159c2d7d05b018051d6e22408359e952a6

corinnekrych:
@jarifibrahim here is https://github.com/jarifibrahim/devopsconsole-operator/pull/1 to fix `make local`

corinnekrych:
@jarifibrahim now i run into:
```
{"level":"error","ts":1553777735.125998,"logger":"kubebuilder.controller","caller":"controller/controller.go:209","msg":"Reconciler error","Controller":"installer-controller","Request":"local-test/main-installer","error":"APP_SERVICE_IMAGE_NAME must be set","stacktrace":"github.com/redhat-developer/devopsconsole-operator/vendor/github.com/go-logr/zapr.(*zapLogger).Error\n\t/Users/corinne/workspace/go/src/github.com/red
```
what do I need to docker build? the service?

jarifibrahim:
> @jarifibrahim now i run into:
> 
> ```
> {"level":"error","ts":1553777735.125998,"logger":"kubebuilder.controller","caller":"controller/controller.go:209","msg":"Reconciler error","Controller":"installer-controller","Request":"local-test/main-installer","error":"APP_SERVICE_IMAGE_NAME must be set","stacktrace":"github.com/redhat-developer/devopsconsole-operator/vendor/github.com/go-logr/zapr.(*zapLogger).Error\n\t/Users/corinne/workspace/go/src/github.com/red
> ```
> what do I need to docker build? the service?

You can build a docker image for https://github.com/redhat-developer/app-service or you can use `jarifibrahim/boilerplate-app` image.
Run 
```
export APP_SERVICE_IMAGE_NAME="jarifibrahim/boilerplate-app"
```
and then 
```
operator-sdk up local
```

deploy/crds/devopsconsole_v1alpha1_installer_cr.yaml
tinakurian: Maybe we can create a separate story for this? It looks like this story has already ballooned from the original description in the story. We should probably try to avoid doing this as it messes up our estimates in terms of how much work we planned the story to be vs how much work it actually turned out to be.

corinnekrych:
- tested on local env, works fine
- haven't tested olm yet

corinnekrych:
@jarifibrahim I've fixed the following RBAC issue` with https://github.com/jarifibrahim/devopsconsole-operator/pull/2:
```
1) E0328 19:07:15.847379       1 reflector.go:205] github.com/redhat-developer/devopsconsole-operator/vendor/sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:126: Failed to list *v1.Deployment: deployments.apps is forbidden: User "system:serviceaccount:devopsconsole:devopsconsole-operator" cannot list deployments.apps in the namespace "devopsconsole": no RBAC policy matched

2) E0328 19:11:40.781946       1 reflector.go:322] github.com/redhat-developer/devopsconsole-operator/vendor/sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:126: Failed to watch *v1.Deployment: unknown (get deployments.apps)

3) E0328 19:11:41.753972       1 reflector.go:205] github.com/redhat-developer/devopsconsole-operator/vendor/sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:126: Failed to list *v1.Route: routes.route.openshift.io is forbidden: User "system:serviceaccount:devopsconsole:devopsconsole-operator
```

corinnekrych:
@jarifibrahim we're left with one issue:
in [operator.yaml]() the namespace is left empty to mean watch all namespace. However with this file, the deployment of installer fails with:
```
{"level":"error","ts":1553801290.8579235,"logger":"cmd","caller":"manager/main.go:143","msg":"failed to create installer resource","error":"an empty namespace may not be set during creation”,"stacktrace":"github.com/redhat-developer/devopsconsole-operator/vendor/github.com/go-logr/zapr.(*zapLogger).Error\n\t/Users/corinne/workspace/go/src/github.com/redhat-developer/devopsconsole-operator/vendor/github.com/go-logr/zapr/zapr.go:128\nmain.main\n\t/Users/corinne/workspace/go/src/github.com/redhat-developer/devopsconsole-operator/cmd/manager/main.go:143\nruntime.main\n\t/Users/corinne/.gvm/gos/go1.11.2/src/runtime/proc.go:201"}
```
If I put back the reference in namespace field the installer works fine BUT the component controller does not watch all namespace.
We need this value empty for our cluster-wide operator. I guess we need to change code in `main.go`. Any idea?

jarifibrahim:
> @jarifibrahim we're left with one issue:
> in operator.yaml the namespace is left empty to mean watch all namespace. However with this file, the deployment of installer fails with:
> 
> ```
> {"level":"error","ts":1553801290.8579235,"logger":"cmd","caller":"manager/main.go:143","msg":"failed to create installer resource","error":"an empty namespace may not be set during creation”,"stacktrace":"github.com/redhat-developer/devopsconsole-operator/vendor/github.com/go-logr/zapr.(*zapLogger).Error\n\t/Users/corinne/workspace/go/src/github.com/redhat-developer/devopsconsole-operator/vendor/github.com/go-logr/zapr/zapr.go:128\nmain.main\n\t/Users/corinne/workspace/go/src/github.com/redhat-developer/devopsconsole-operator/cmd/manager/main.go:143\nruntime.main\n\t/Users/corinne/.gvm/gos/go1.11.2/src/runtime/proc.go:201"}
> ```
> If I put back the reference in namespace field the installer works fine BUT the component controller does not watch all namespace.
> We need this value empty for our cluster-wide operator. I guess we need to change code in `main.go`. Any idea?

@corinnekrych Somehow I didn't face this issue. I have tried doing the entire OLM flow a couple of time but I never faced this issue. 

Are you running the operator locally?

corinnekrych:
@jarifibrahim to reproduce the issue:
- either follow this https://github.com/redhat-developer/devconsole-operator#deploy-the-operator-with-deployment-yaml
- alsoI think the change i've suggested here is wrong: https://github.com/redhat-developer/devconsole-operator/pull/31/files#diff-b67911656ef5d18c4ae36cb6741b7965R290 keep this empty and you'll see the issue jus with a `make local`

jarifibrahim:
> @jarifibrahim to reproduce the issue:
> 
> * either follow this https://github.com/redhat-developer/devconsole-operator#deploy-the-operator-with-deployment-yaml
> * alsoI think the change i've suggested here is wrong: https://github.com/redhat-developer/devconsole-operator/pull/31/files#diff-b67911656ef5d18c4ae36cb6741b7965R290 keep this empty and you'll see the issue jus with a `make local`

Yeah. I understood that but I'm not sure what would be the way forward. The `installer` CR has to be created in a namespace. Which namespace should that be? How do we figure out the namespace?

corinnekrych:
We moved the component operator to be cluster-wide to be able to watch all namespace (is a cluster resource).
Similarly the operator is installed in one "operator" namespace but watch for all namescape. ie https://github.com/redhat-developer/devconsole-operator/blob/master/deploy/operator.yaml#L25

But your code here https://github.com/redhat-developer/devconsole-operator/pull/31/files#diff-9e63f7b0791edcb4a53c944552b88970R140 need to take not the watched namespace but the one where you've installed the operator.

baijum:
> Let's rebase and merge

I think we should merge upstream/master and not rebase.

deploy/operator.yaml
jarifibrahim: Updated the image.

PR #39, opened 2019-03-22
updated .gitignore with gitignore.io snippet

kwk:
With https://www.gitignore.io/?templates=go,vim,git,macos,linux,emacs,eclipse,windows,visualstudiocode we have a nice and easy way to manage out git ignore file.

The pieces that are relevant just for this project appear at the top of the `.gitignore` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,146 @@
-### IntelliJ ###
-.idea/
-*.iws
-*.iml
-/out/
-.idea_modules/
+out/
+coverage.txt
+bin/
+tmp/
+
+# Created by https://www.gitignore.io/api/go,vim,git,macos,linux,emacs,eclipse,windows,visualstudiocode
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,9 +1,146 @@
-### IntelliJ ###
-.idea/
```
corinnekrych: @kwk would you mind leaving the `.idea/` for the goland IDE in the `.gitignore`
I see you adding other IDE which is fine but let's leave that one

kwk: Why do we want to put any IDE setting under version control? I don't see why we should do that?

corinnekrych: No
Like you ignore eclipse, emac, ignore also goland idea

kwk: Ah :) sorry, I misread the comment as: "please add it". On it.

kwk: I've updated the gitignore.io URL to include intellij: https://www.gitignore.io/?templates=go,vim,git,macos,linux,emacs,windows,eclipse,intellij+all,visualstudiocode Are you fine with this @corinnekrych ?

PR #42, opened 2019-03-24
Makefile target to build image using operator SDK

baijum:
Resolves: https://jira.coreos.com/browse/ODC-357

### DIFF
--- a/make/Makefile.dev
+++ b/make/Makefile.dev
@@ -5,6 +5,7 @@ ifneq ($(GITUNTRACKEDCHANGES),)
 SHORT_COMMIT := $(SHORT_COMMIT)-dirty
 endif
 
+DEVOPSCONSOLE_IMAGE?=quay.io/redhat-developers/devopsconsole-operator
 TIMESTAMP:=$(shell date +%s)
 TAG?=$(SHORT_COMMIT)-$(TIMESTAMP)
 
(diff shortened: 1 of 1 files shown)


make/Makefile.dev
```diff
@@ -5,6 +5,7 @@ ifneq ($(GITUNTRACKEDCHANGES),)
 SHORT_COMMIT := $(SHORT_COMMIT)-dirty
 endif
 
+DEVOPSCONSOLE_IMAGE?=quay.io/redhat-developers/devopsconsole-operator
```
baijum: This can be updated through the environment variable `DEVOPSCONSOLE_IMAGE`.

baijum:
There are multiple images part of the Developer Console:

- The Developer Console operator image
- The operator registry image
- Multiple service images
- Multiple dependent operator images

A naming convention should be created to denote these images.

The top-level operator image could be `DEVOPSCONSOLE_OPERATOR_IMAGE`.
Later it can be changed to `DEVCONSOLE_OPERATOR_IMAGE` once we finalize
on the name.

The operator registry is going to contain CSVs of the top-level operator
and dependent operators.  Since there are no dependent operators, it's
going to contain contain the top-level image alone.  The operator
registry image could be `DEVOPSCONSOLE_OPERATOR_REGISTRY_IMAGE`.

The service images can follow a naming pattern like this:
`<service-name>_SERVICE_IMAGE`.  If the should be further qualified, add
a prefix of `DEVOPSCONSOLE_`.

The dependent operator images could be: `<operator-name>_OPERATOR_IMAGE`.

PR #43, opened 2019-03-25
Pushing the operator image to a container registry

baijum:
Resolves: https://jira.coreos.com/browse/ODC-358

### DIFF
--- a/make/Makefile.dev
+++ b/make/Makefile.dev
@@ -38,6 +38,12 @@ create-cr:
 build-operator-image:
 	operator-sdk build $(DEVOPSCONSOLE_OPERATOR_IMAGE)
 
+.PHONY: push-operator-image
+## Push the operator container image to a container registry
+push-operator-image: build-operator-image
+	docker login -u $(QUAY_USERNAME) -p $(QUAY_PASSWORD) "quay.io"
+	docker push $(DEVOPSCONSOLE_OPERATOR_REGISTRY)
+
 .PHONY: deploy-operator-only
 deploy-operator-only:
 	@echo "Creating Deployment for Operator"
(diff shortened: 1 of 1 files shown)


make/Makefile.dev
```diff
@@ -38,6 +38,12 @@ create-cr:
 build-operator-image:
 	operator-sdk build $(DEVOPSCONSOLE_OPERATOR_IMAGE)
 
+.PHONY: push-operator-image
+## Push the operator container image to a container registry
+push-operator-image: build-operator-image
+	docker login -u $(QUAY_USERNAME) -p $(QUAY_PASSWORD) "quay.io"
```
corinnekrych: Maybe reuse `REGISTRY_URI = quay.io` from  https://github.com/redhat-developer/devopsconsole-operator/blob/master/Makefile#L60 
We've got a bit of duplication between those 2 makefiles

make/Makefile.dev
```diff
@@ -38,6 +38,12 @@ create-cr:
...
+	docker push $(DEVOPSCONSOLE_OPERATOR_REGISTRY)
```
corinnekrych: @baijum Is your PR dependent on another PR? where is DEVOPSCONSOLE_OPERATOR_REGISTRY defined?

baijum: I had another PR which introduced it. I merged today morning.

baijum: Fixed 50c2e205deee1acbb6e505943c9f4743443d3a29

make/Makefile.dev
baijum: Done. 28d030818095cea6383d25c31dfccef38b39cfde

PR #44, opened 2019-03-25
add deployment config creation in component S2I flow

corinnekrych:
Fixes https://jira.coreos.com/browse/ODC-67

Similar to the PoC PR https://github.com/redhat-developer/devopsconsole-operator/pull/6/files finish the S2I flow with DeploymentConfig creation.


corinnekrych:
@sbose78 I took the definition of dc in your PR, but i'll double checked with the one used in odo which triggers automatic deployment.

pkg/controller/component/component_controller.go
```diff
@@ -282,3 +299,56 @@ func generateBuildConfig(namespace string, name string, builder *builderImage, g
 		},
 	}
 }
+
+func generateDeploymentConfig(namespace string, name string) *v1.DeploymentConfig {
+	labels := map[string]string{
+		"app": name,
+	}
+	return &v1.DeploymentConfig{
```
sbose78: How about creating a Deployment instead of a DeploymentConfig? I'm saying this because DeploymentConfigs don't exist in Kubernetes.

pkg/controller/component/component_controller.go
```diff
@@ -282,3 +299,56 @@ func generateBuildConfig(namespace string, name string, builder *builderImage, g
...
+			Template: &corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      name,
+					Namespace: namespace,
+					Labels:    labels,
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{{
+						Name:  name,
+						Image: name + "-output:latest",
```
sbose78: I know we have this everywhere, but do we really need the name suffixes?

pkg/controller/component/component_controller.go
```diff
@@ -282,3 +299,56 @@ func generateBuildConfig(namespace string, name string, builder *builderImage, g
...
 	}
 }
+
+func generateDeploymentConfig(namespace string, name string) *v1.DeploymentConfig {
+	labels := map[string]string{
+		"app": name,
+	}
+	return &v1.DeploymentConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name + "-dc",
```
sbose78: Do we really need the suffix here?

sbose78: Reason I ask is: oc new-app probably doesn't do it.

pkg/controller/component/component_controller.go
corinnekrych: agreed, prefix is useless, label selectors should be used instead. I'll remove.

sbose78: Cool, thank you.

pkg/controller/component/component_controller.go
corinnekrych: done as part of f1e940c

PR #46, opened 2019-03-27
WIP add gitpush CRD

sbryzak:
This PR adds the git-push custom resource, used to represent a git push operation.  

**TO DISCUSS** - there remains a question of naming.. should we be using `devconsole.openshift.io`, or `devopsconsole.openshift.io`?  The impression from mailing list discussions are that we are going with "Developer Console", making `devconsole.openshift.io` the more likely package naming.


sbose78:
Let's use `devconsole.openshift.io`

PR #34, opened 2019-03-21
setting up a basic e2e test to verify component CR get created

corinnekrych:
fixed odc-132
This PR adds:
- a simple e2e test for deploying `devopsconsole-operator` and creating a component CR component
- add Makefile support for testing locally in dev mode (test are not run in container) for easy writing and debugging of the 2e2 tests

PR #47, opened 2019-03-27
Cleanup Makefile and Dockerfile 

kwk:
## Makefile

By looking at our Makefiles we can identify several topics:

- building
- testing
- containerization
- help
- sanity checks (e.g. linting and formatting)
- ...

I've isolate each of those topics into their own makefiles under `./make` and and tried to reduce the number of common dependencies to a minimum. For example, if your Makefile needs to know something about make the git version, just include

```Makefile
include ./make/git.mk
```
**NOTE:** You can include each `./make/*.mk` as many times as you want. They all have include guards and are only included once.

## Dockerfile

I wanted for quay.io or docker hub to be able to take and build our code as easily as possible and that means we need to have a single `Dockerfile`. Thanks to multi-stage Dockerfiles, we don't loose any of our benefits that we did have with our `Dockerfile.builder` for example.

I've unified our Dockerfiles into one `Dockerfile` with three stages:

1. The first stage is called _build-tools_ and gives us a centos:7 image with all tools installed to build the software. No build or test is run in this stage.
2. The second stage, aka _builder_, is based upon the _build-tools_ stage and indeed builds the software and tests it (without e2e tests). 
3. The last and third stage, aka _deploy_, is again based on centos:7 and nothing else. None of the tools from _build_tools_ are available here. Then we copy over the build artifacts from the _builder_ stage to the _deploy_ stage.

## Artifacts

We nevery truely supported out-of-source builds and it is quite uncommon to do so with Go projects. That's why I've decided to simplify the makefiles a lot. You'll no longer find things like `$(CUR_DIR)/somefile`, but instead you'll see `./somefile`. The assumption is simply that you run `make` from the root directory of the project.

### ./out 

As a side-effect I decided to remove some more path variables as well. For example if your targets needs to write an artifact and location doesn't matter for you that much, you can safely assume that the directory `./out` is there. It will be cleaned up upon `make clean` for you, so you don't have to add a special clean target for your artifact.  

## Test simplification and cross-package coverage

As of go 1.9 `go list ./...` will exclude the vendor directory, so there's no need for us to filter out the vendor directory. Most of the test code that we had before was written by me in the github.com/fabric8-services/fabric8-wit repository and it was blindly copied over here without knowing what it does. I'm not particularly proud for the test code in fabric8-wit but back then it was the only way we could achieve _cross-package coverage_.

The idea for _cross-package coverage_ was to collect and merge the coverage information from package `a` and `b` if they call functions in tests from one another. Theoretically this sounds like a good idea. But consider that you want to take package `b` and use it somewhere else. Then all of a sudden the package coverage can drop drastically if most of the testing for `b` was done through package `a`.

I want a package to test it's code as best as it can and nothing more.

The result is what you can see in `./make/test.mk`. Testing the Go code can finally be fun again and if you want to take a look at the coverage of the project, simply run `make test-coverage` (for terminal coverage reports) or `make test-coverage-html` (for viewing line-by-line reports in your browser).  

## Linting YAML and Go

With the new `make lint-yaml` you can run `yamllint` on all non-vendor YAML files. 

I've made my life easy by simply `go get`ting the golang-ci instead of downloading a special build for each architecture and checking it's checksum before installing it. Once Go 1.13 is there we can `go get` a specific version.

## Finding tools

Instead of having a `prebuild` target in our makefile we're now looking for all the required tools on each invocation of `make`. The only exception is when you run `make help` or `make docker-XXX`. For `make help` we check for no tool at all. For `make docker-XXX` we just check for the `docker` executable. See the `./make/find-tools.mk` file for more information.

**NOTE:** We don't control which version each tool has as the moment.

## Verbosity

I wanted to be able to configure the output of `make` in these ways:

1. `make VERBOSE=0 <TARGET>`: Reduce output to minimum (not even showing the command that is executed).
2. `make VERBOSE=1 <TARGET>` (the default): Show the command that is executed but don't turn on `-v` verbosity flags for each command.
3. `make VERBOSE=2 <TARGET>`: Show the command as in 2. but also turn on all verbosity switches and disable silent flags like `-s`.

Have a look at the usage of variables like `${V_FLAG}` and `${S_FLAG}` to find out how to use them.

By prefixing a command with `$(Q)` instead of `@` we can control if the command shall be printed before executing. `$(Q)` expands to `@` when `make` is called with `VERBOSE=0` and to `` (empty) when called with any other number.


cmd/manager/main.go
```diff
@@ -23,6 +25,17 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/runtime/signals"
 )
 
+var (
+	// Commit current build commit set by build script
+	Commit = "0"
```
kwk: This is not used at the moment. I wonder if we need it at all.

Makefile
```diff
@@ -1,323 +1,64 @@
...
+	$(Q)go clean ${X_FLAG} ./...
+
+GIT_COMMIT_ID := $(shell git rev-parse HEAD)
+ifneq ($(shell git status --porcelain --untracked-files=no),)
+       GIT_COMMIT_ID := $(GIT_COMMIT_ID)-dirty
+endif
+BUILD_TIME = `date -u '+%Y-%m-%dT%H:%M:%SZ'`
+
+./out/operator: ./vendor $(shell find . -path ./vendor -prune -o -name '*.go' -print)
+	$(Q)operator-sdk generate k8s
```
sbose78: Can we remove this, please? I don't wish to be generating 'critical' code at build time.

kwk: I did ask you and @corinnekrych  that yesterday and I did get the impression that we have to run this command if the CRD changes. We did the same for fabric8-wit and the design/controller files, no?

sbose78: I had said that actually said the opposite. The operator-sdk always doesn't output reliable content since it's also a development in progress - which is why I prefer a developer to generate it locally, vet it and commit it separately.

kwk: Okay, sorry for the misunderstanding. I've commented out the generation in 98c35bc.

ldimaggi:
I like the approach - smaller, narrow focused makefiles - should be easier to maintain. I was wondering about the single Dockerfile - but iot's good as we can generate multiple images from it.

make/dev.mk
```diff
@@ -1,37 +1,37 @@
...
-	@oc create -f $(DEPLOY_DIR)/role.yaml
-	@echo "Creating RoleBinding"
-	@oc create -f $(DEPLOY_DIR)/role_binding.yaml
+	$(info Logging using system:admin...)
+	$(Q)oc login -u system:admin
+	$(info Creating sub resources...)
+	$(info Creating CRDs...)
+	$(Q)oc create -f ./deploy/crds/devopsconsole_v1alpha1_gitsource_crd.yaml
+	$(info Creating Namespace)
+	$(Q)oc create -f ./deploy/namespace.yaml
```
tinakurian: I think we were trying to avoid using relative paths - please see the following comment: https://github.com/redhat-developer/devconsole-operator/pull/34#discussion_r268008097

kwk:
There is no reason to not use relative paths as I‘ve outlined in the description of the PR. We are not building out of source, so all is good.

make/dev.mk
kwk: @tinakurian There is no reason to not use relative paths as I‘ve outlined in the description of the PR. We are not building out of source, so all is good.

corinnekrych:
@kwk my feedback and minor fix for macOs here: https://github.com/kwk/devopsconsole-operator/pull/1

sbose78:
Konrad, are you addressing Jessica's recommendation ?


> recommendation for new operators is to use the universal base image (UBI) which is registry.access.redhat.com/ubi7/ubi-minimal:latest  by using UBI you get to use the same base for upstream and downstream builds of your operator.
> 
> The hard requirement is that it must be either a UBI or RHEL base that goes into the product pipeline (into the redhat-operators source for the marketplace).

kwk:
> Konrad, are you addressing Jessica's recommendation ?
> 
> > recommendation for new operators is to use the universal base image (UBI) which is registry.access.redhat.com/ubi7/ubi-minimal:latest  by using UBI you get to use the same base for upstream and downstream builds of your operator.
> > The hard requirement is that it must be either a UBI or RHEL base that goes into the product pipeline (into the redhat-operators source for the marketplace).

@sbose78 I first need to find out which list you're talking about here: https://chat.openshift.io/developers/pl/ms6jcpn3bjbfxe31q1wgktj1zc

kwk:
@sbose78 also my question would be if we need to **build** AND **deploy** on UBI or RHEL or if we only need to **deploy** with those? To formulate things differently. Can we build on centos:7 and deploy with UBI or RHEL? Just asking because I don't know if UBI or RHEL can build it. Will try now.

kwk:
@sbose78 so how is quay.io going to build our images or will it just get pushed whatever we've created? I'm asking because I cannot pull the UBI:

```
$ docker pull registry.access.redhat.com/ubi7/ubi-minimal:latest
Error response from daemon: error parsing HTTP 404 response body: invalid character 'F' looking for beginning of value: "File not found.\""
```

tinakurian:
> @sbose78 so how is quay.io going to build our images or will it just get pushed whatever we've created? I'm asking because I cannot pull the UBI:
> 
> ```
> $ docker pull registry.access.redhat.com/ubi7/ubi-minimal:latest
> Error response from daemon: error parsing HTTP 404 response body: invalid character 'F' looking for beginning of value: "File not found.\""
> ```
> There is no result for UBI here as well: https://access.redhat.com/containers/#/product/5c180b28bed8bd75a2c29a63 It says something about UBI but other than that: no information available.

I don't see this either. I do, however see rhel atomic base image..

sbose78:
>> The location of the image has moved. Try: `registry.access.redhat.com/ubi7-dev-preview/ubi-minimal:latest` . 

Regarding - 

>>  Can we build on centos:7 and deploy with UBI or RHEL? Just asking because I don't know if UBI or RHEL can build it. Will try now.

I'm not sure if it would build either - worth exploring though.
Ideally, building using UBI would be nice too. But not a must-have as far as I can tell because Golang anyway produces an executable which could then be run using a UBI base. 

I will merge this PR so I can consume this as part of my openshift CI PR - you could handle UBI usage in a separate PR.

Dockerfile
```diff
@@ -0,0 +1,86 @@
...
+RUN curl -L -s https://github.com/openshift/origin/releases/download/v3.11.0/openshift-origin-client-tools-v3.11.0-0cbc58b-linux-64bit.tar.gz -o openshift-origin-client-tools.tar.gz \
+    && echo "4b0f07428ba854174c58d2e38287e5402964c9a9355f6c359d1242efd0990da3 openshift-origin-client-tools.tar.gz" > openshift-origin-client-tools.sha256 \
+    && sha256sum -c openshift-origin-client-tools.sha256 \
+    && tar xzf openshift-origin-client-tools.tar.gz \
+    && mv /tmp/openshift-origin-client-tools-*/oc /usr/bin/oc \
+    && mv /tmp/openshift-origin-client-tools-*/kubectl /usr/bin/kubectl \
+    && rm -rf ./openshift* \
+    && oc version
+
+# install operator-sdk (from git with no history and only the master branch)
```
sbose78: I will revisit if we can do this differently. I wish to avoid 
1. operator-sdk as a build time dependency
2. using operator-sdk master.

Makefile
```diff
@@ -1,320 +1,37 @@
...
+clean:
+	$(Q)-rm -rf ${V_FLAG} ./out
+	$(Q)-rm -rf ${V_FLAG} ./vendor
+	$(Q)go clean ${X_FLAG} ./...
+
+./vendor: Gopkg.toml Gopkg.lock
+	$(Q)dep ensure ${V_FLAG} -vendor-only
+
+./out/operator: ./vendor $(shell find . -path ./vendor -prune -o -name '*.go' -print)
+	#$(Q)operator-sdk generate k8s
```
sbose78: Fine to have the operator-sdk as a built time dependency since we are not doing a `generate k8s`.

Dockerfile
kwk: @sbose78 I want to keep it because then you can add make targets that use the operator-sdk and the build-tools image is ready. Also to switch to some version more specific, just checkout whatever ref you want.

Issue #40, opened 2019-03-22
Ensure latest CRDs under manifests directory and deploy/crds directory are same


baijum:
Write a test to ensure the latest CRDs under manifests directory and deploy/crds directory are same

PR #48, opened 2019-03-27
Check for the sanity of CRD files

baijum:
Resolves: #40

The latest CRD files under the manifests directory and the CRD
files under deploy/crds should match.


corinnekrych:
@baijum I'm really buying into this sanity check until we got sth even better like only one source of truth. 😍 

I think it would be good to add another check (more work since it's not just file verification) to check the content of `deploy/role.yaml` and compare it to `clusterPermission` section of `devopsconsole-operator.v0.1.0.clusterserviceversion.yaml`

mmm since we'll got on per version, long run we need some sort of code generation.

corinnekrych:
@baijum I've started a discussion in here: https://coreos.slack.com/archives/C3VS0LV41/p1553697141394800
and https://coreos.slack.com/archives/C3VS0LV41/p1553697786396900?thread_ts=1553697687.396600&cid=C3VS0LV41
and it seems we got:
```
operator-sdk olm-catalog gen-csv --csv-version 0.2.0`
```
that will generate `deploy\devo[sconsole-oprator.csv.yaml` and `deploy/_generated.concat_crd.yaml`
So I think a better approach will be to have those files generated as part of the build. wdyt?

baijum:
> @baijum I've started a discussion in here: https://coreos.slack.com/archives/C3VS0LV41/p1553697141394800
> and https://coreos.slack.com/archives/C3VS0LV41/p1553697786396900?thread_ts=1553697687.396600&cid=C3VS0LV41

Thanks for bringing up this into the operator framework developers.

> and it seems we got:
> 
> ```
> operator-sdk olm-catalog gen-csv --csv-version 0.2.0`
> ```
> that will generate `deploy\devo[sconsole-oprator.csv.yaml` and `deploy/_generated.concat_crd.yaml`
> So I think a better approach will be to have those files generated as part of the build. wdyt?

Once we have good test coverage, we can evaluate those options.

test/sanitychecks/crd_test.go
```diff
@@ -0,0 +1,64 @@
...
+	Channels []struct {
+		CurrentCSV string `yaml:"currentCSV"`
+		Name       string `yaml:"name"`
+	} `yaml:"channels"`
+	PackageName string `yaml:"packageName"`
+}
+
+func TestLatestCRDFiles(t *testing.T) {
+	curdir, err := os.Getwd()
+	if err != nil {
```
kwk: What about `require.NoError(t, err)`?

baijum: Done. 6a786427ccd6fafb9eefa7a480182b7830135842

PR #52, opened 2019-03-28
Test all the text file ends with a newline at EOF

baijum:
Ref.
https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline


akashshinde:
@baijum Instead of running test to ensure this, can't we add pre-commit hook which would add new line if it's not present ? https://githooks.com/

baijum:
> @baijum Instead of running test to ensure this, can't we add pre-commit hook which would ensure to have extra line or adds new line if it's not present ? https://githooks.com/

Is this can be set up in CI?

akashshinde:
> > @baijum Instead of running test to ensure this, can't we add pre-commit hook which would ensure to have extra line or adds new line if it's not present ? https://githooks.com/
> 
> Is this can be set up in CI?

I think we wouldn't need to setup extra step in CI since commit pre-hook script would always get executed before anyone commits in their local repo. It wouldn't allow you to commit until and unless you fix it.

baijum:
Akash, What if someone forget to setup the hook? Is this can be automated, so that everyone gets the same hook. Is this requires some conscious effort from everyone from the team?

PR #62, opened 2019-04-01
Rename devops to dev in remaining files

baijum:


baijum:
Thanks @corinnekrych !

PR #55, opened 2019-03-28
Merging pkg/devopsconsole-operator into pkg/devopsconsole

tinakurian:
Should contain a test for gitsource - however there is some test duplication that needs to be sorted out: https://github.com/redhat-developer/devconsole-operator/pull/33/files#r269921383

Will add the gitsource test once the test duplication has been resolved.

PR #33, opened 2019-03-21
Create Makefile target to install operator & test

Avni-Sharma:
Refer: https://jira.coreos.com/browse/ODC-189

This pull request introduces a new Makefile target to run the Operator Lifecycle Manager (OLM) integration test. The target is `test-olm-integration`, and you can run it like this:

```
make test-olm-integration
```

Before running the above, you should set environment variables listed here:

```
export QUAY_USERNAME=
export QUAY_PASSWORD=
export DEVCONSOLE_OPERATOR_IMAGE=quay.io/<username>/devconsole-operator
export DEVCONSOLE_OPERATOR_REGISTRY_IMAGE=quay.io/<username>/operator-registry
```

The `QUAY_USERNAME` and `QUAY_PASSWORD` is used to login to Quay.  The `DEVCONSOLE_OPERATOR_IMAGE` environment variable is pointing to the operator image location without the tag. The`DEVCONSOLE_OPERATOR_REGISTRY_IMAGE` is pointing to the registry image without the tag.

PR #67, opened 2019-04-02
Fix e2e local

sbose78:
- rename repetitive code which caused compile time issues in the end
- add runtime env var which was used to determine `if Darwin`


test/e2e/component_test.go
```diff
@@ -18,13 +17,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-const (
-	retryInterval        = time.Second * 5
```
baijum: I don't see these constants defined anywhere else. And it is used in `TestComponent`.

sbose78: https://github.com/redhat-developer/devconsole-operator/blob/master/test/e2e/custom_resource_operator_test.go#L25

baijum: That file was removed in Avni's PR #33 But looks like it got introduced again from PR #55

baijum: We should remove that file, and just keep `component_test.go`. It is the same test.

tinakurian: Fixed

sbose78: AH yes!

PR #68, opened 2019-04-02
Clean up of catalog source and subscription

### LINKED ISSUE #65: Consider cleaning up catalogSource, Subscription
.. after e2e tests.

https://github.com/redhat-developer/devconsole-operator/pull/33#issuecomment-478650933

Avni-Sharma:
Fixes #65


make/test.mk
```diff
@@ -80,6 +80,8 @@ olm-integration-setup: olm-integration-cleanup
 .PHONY: olm-integration-cleanup
 olm-integration-cleanup: get-test-namespace
 	$(Q)oc login -u system:admin
+	$(Q)-oc delete catalogsources -n olm my-catalog
```
baijum: The above code should work. However, writing it like this would be better.
The syntax should be `$(Q)-oc delete catalogsource my-catalog -n olm`

(Same with subscription also)

baijum:
Probably we should also try to delete clusterrole and clusterrolebindings (both has the name as `devconsole-operator`)

sbose78:
yeah, something like `oc delete -f deploy/cluster_roles.yaml..` ?

baijum:
The role and role bindings are created by OLM based on the manifests, so the test framework should clean up that. But sometimes it doesn't work. So, it's an issue of the Operator SDK test framework. We need to follow up with them.

For now catalog source & subscription cleanup is fine.

PR #72, opened 2019-04-03
Add Dockerfile which has image with binary

sbose78:
.. for openshift ci


kwk:
@sbose78 why do you do that? We could as well just use [this line](https://github.com/redhat-developer/devconsole-operator/blob/2ba01e0c875e5616479f3da5b0f88edf0743e51a/Dockerfile#L81) that does the exact same thing:

```Dockerfile
COPY --from=builder ${GOPATH}/src/${GO_PACKAGE_PATH}/out/operator /usr/local/bin/devconsole-operator
```

PR #74, opened 2019-04-03
Correct log message

DhritiShikhar:


akashshinde:
@DhritiShikhar  @baijum  Shouldn't it be **_operator is running_** ?

baijum:
> @DhritiShikhar @baijum Shouldn't it be **_operator is running_** ?

That message with the one Dhriti corrected would be better,
Something like this:
`devconsole-operator is running and component CRD is ready for use.`

PR #76, opened 2019-04-03
after deployment add service creation

corinnekrych:
fixes https://jira.coreos.com/browse/ODC-382


examples/devconsole_v1alpha1_component_cr.yaml
```diff
@@ -5,4 +5,6 @@ metadata:
 spec:
   buildType: "nodejs"
   codebase: "https://github.com/sclorg/nodejs-ex" #"https://github.com/nodeshift-starters/nodejs-rest-http-crud"
+  port: "8080"
```
dipak-pawar: we are using `8080` as default, maybe we can some different value, to not confuse user who is reading this example

manifests/devconsole/0.1.0/devconsole_v1alpha1_component_crd.yaml
```diff
@@ -31,6 +31,13 @@ spec:
               - nodejs
             codebase:
               type: string
+              description: The git url of your component source code.
+            port:
+              type: string
```
dipak-pawar: How about type `integer` and having some min and max limit - `65635`?

pkg/controller/component/component_controller.go
```diff
@@ -128,12 +129,85 @@ func (r *ReconcileComponent) Reconcile(request reconcile.Request) (reconcile.Res
...
+			return reconcile.Result{}, err
+		}
+		if instance.Spec.Exposed == true {
+			_, err = r.CreateRoute(instance)
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+		}
+
+		log.Info("** ImageStream, BuildConfig, DeploymentConfig and Service successfully created. **")
```
dipak-pawar: how about adding Route here when `--exposed` is true and it got created.

pkg/controller/component/component_resources.go
```diff
@@ -142,3 +145,66 @@ func newDeploymentConfig(cr *componentsv1alpha1.Component, output *imagev1.Image
...
+			Namespace: cr.Namespace,
+			Labels:    labels,
+		},
+		Spec: routev1.RouteSpec{
+			To: routev1.RouteTargetReference{
+				Kind: "Service",
+				Name: cr.Name,
+			},
+			Port: &routev1.RoutePort{
+				TargetPort: intstr.FromInt(int(int32Port)),
```
dipak-pawar: while trying `oc new-app` I have see when you expose route you see following
```
apiVersion: route.openshift.io/v1
kind: Route
metadata:
  annotations:
    openshift.io/host.generated: "true"
  creationTimestamp: 2019-04-03T11:12:34Z
  labels:
    app: ruby-ex
  name: ruby-ex
  namespace: test
  resourceVersion: "33511"
  selfLink: /apis/route.openshift.io/v1/namespaces/test/routes/ruby-ex
  uid: 6196ac77-5601-11e9-a8ab-525400e16554
spec:
  host: ruby-ex-test.192.168.42.138.nip.io
  port:
    targetPort: 8080-tcp
  to:
    kind: Service
    name: ruby-ex
    weight: 100
  wildcardPolicy: None
```
Just wondering why `TargetPort` does not have `tcp` appended, as we have port name appended with `-tcp` in service. see https://github.com/redhat-developer/devconsole-operator/pull/76/files#diff-3240f219b66b3cb310b8835fe5cf7e42R165

pkg/controller/component/component_controller_test.go
```diff
@@ -123,6 +130,100 @@ func TestComponentController(t *testing.T) {
...
+		_, err := r.Reconcile(req)
+
+		//then
+		require.NoError(t, err, "reconcile is failing")
+
+		svc := &corev1.Service{}
+		errGetSvc := cl.Get(context.Background(), types.NamespacedName{Namespace: Namespace, Name: Name}, svc)
+		require.NoError(t, errGetSvc, "service is not created")
+		require.Equal(t, 1, len(svc.Spec.Ports), "service is using one port")
+		require.Equal(t, int32(3000), svc.Spec.Ports[0].Port, "service port should be 3000")
```
dipak-pawar: I would also verify endpoint for that service is created. you can find that using `oc get ep`

manifests/devconsole/0.1.0/devconsole-operator.v0.1.0.clusterserviceversion.yaml
```diff
@@ -135,6 +135,15 @@ spec:
...
           - watch
+        - apiGroups:
+          - route.openshift.io
+          resources:
+          - routes
+          verbs:
+          - get
+          - create
+          - list
+          - watch
```
dipak-pawar: just wondering, we are not watching any routes in this pr, do we still needs it?

dipak-pawar:
@corinnekrych Looks like it is approve by you. I am not sure how PR owner/creator can approve it's own PR.

sbose78:
@dipak-pawar if the one who opened the PR is in the OWNER list, prow labels it as approved.
I need to see if I can avoid that, but I see this behaviour in the openshift/release repo as well

manifests/devconsole/0.1.0/devconsole-operator.v0.1.0.clusterserviceversion.yaml
corinnekrych: yes, otherwise I bumped into an rbac issue.

pkg/controller/component/component_resources.go
corinnekrych: 425b8f4

manifests/devconsole/0.1.0/devconsole_v1alpha1_component_crd.yaml
corinnekrych: good idea, implemented 425b8f4 with a range [1024-65535]

examples/devconsole_v1alpha1_component_cr.yaml
corinnekrych: ok, changed by 3000

pkg/controller/component/component_controller_test.go
```diff
@@ -46,6 +49,10 @@ func TestComponentController(t *testing.T) {
 	s.AddKnownTypes(compv1alpha1.SchemeGroupVersion, cp)
 
 	// register openshift resource specific schema
+	if err := routev1.AddToScheme(s); err != nil {
+		log.Error(err, "")
+		assert.Nil(t, err, "adding imagestream schema is failing")
```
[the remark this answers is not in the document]
corinnekrych: fixed

corinnekrych:
@dipak-pawar I left a todo, I was trying to add this be4e597 to query for endpoints without success :(

Do you know how to query for those endpoints? I've noticed you can only do:
`oc get endpoints` so I think we do need to use `List`

PR #79, opened 2019-04-03
use the correct namespace depending on target

tinakurian:
Currently the component_test waits for deployment on the "operators" namespace. 

This namespace is only created via the` make test-olm-integration`

However, when running the `make e2e-local target`, "operators" namespace is never created and hence the test will fail. 

This change sets the deployed_namespace var which can be used in the component_test in the waitForDeployment call.

### DIFF
--- a/test/e2e/component_test.go
+++ b/test/e2e/component_test.go
@@ -3,6 +3,7 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"os"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


test/e2e/component_test.go
```diff
@@ -53,8 +54,9 @@ func TestComponent(t *testing.T) {
 	// get global framework variables
 	f := framework.Global
 	t.Log(fmt.Sprintf("namespace: %s", namespace))
+
 	// wait for component-operator to be ready
-	err = e2eutil.WaitForDeployment(t, f.KubeClient, "operators", "devconsole-operator", 1, retryInterval, timeout*2)
+	err = e2eutil.WaitForDeployment(t, f.KubeClient, os.Getenv("DEPLOYED_NAMESAPCE"), "devconsole-operator", 1, retryInterval, timeout*2)
```
corinnekrych: small spelling mistake @tinakurian

corinnekrych:
🎉 @tinakurian great, testing locally e2e works again, thanks for the fix

PR #83, opened 2019-04-04
Proper Cleanup for test-olm-integration to pass

Avni-Sharma:
The makefile target test-olm-integration was failing due to the improper/partial cleanup of some resources by the e2e test framework. This fixes the test by cleaning up ClusterRole and ClusterRoleBinding.

https://github.com/operator-framework/operator-sdk/issues/1287

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -75,8 +75,9 @@ test-olm-integration: push-operator-image olm-integration-setup
 	$(Q)sed -e "s,REPLACE_IMAGE,$(DEVCONSOLE_OPERATOR_REGISTRY_IMAGE):$(devconsole_version)-$(TAG)," ./test/e2e/catalog_source.yaml | oc apply -f -
 	$(Q)oc apply -f ./test/e2e/subscription.yaml
 
-	@-oc delete clusterroles.rbac.authorization.k8s.io "devconsole-operator"
-	@-oc delete clusterrolebindings.rbac.authorization.k8s.io "devconsole-operator"
+	# The following cleanup is required due to a potential bug in the test framework.
+	$(Q)-oc delete clusterroles.rbac.authorization.k8s.io "devconsole-operator"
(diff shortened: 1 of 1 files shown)


make/test.mk
```diff
@@ -75,6 +75,8 @@ test-olm-integration: push-operator-image olm-integration-setup
 	$(Q)sed -e "s,REPLACE_IMAGE,$(DEVCONSOLE_OPERATOR_REGISTRY_IMAGE):$(devconsole_version)-$(TAG)," ./test/e2e/catalog_source.yaml | oc apply -f -
 	$(Q)oc apply -f ./test/e2e/subscription.yaml
 
+	@-oc delete clusterroles.rbac.authorization.k8s.io "devconsole-operator"
```
baijum: Use `$(Q)-` wrapper here and next line.
Also add a comment says that this required due to a potential bug in the test framework.

sbose78:
Apart from @baijum 's comment, looks good. Thank you @Avni-Sharma .

make/test.mk
```diff
@@ -75,6 +75,9 @@ test-olm-integration: push-operator-image olm-integration-setup
 	$(Q)sed -e "s,REPLACE_IMAGE,$(DEVCONSOLE_OPERATOR_REGISTRY_IMAGE):$(devconsole_version)-$(TAG)," ./test/e2e/catalog_source.yaml | oc apply -f -
 	$(Q)oc apply -f ./test/e2e/subscription.yaml
 
+	# The following cleanup is required due to a potential bug in the test framework.
```
baijum: The clean up should go under `olm-integration-cleanup` target. I missed it earlier as I reviewed this from my mobile ;-)

sbose78: That target is called at the end or start of the test right?

baijum: start of the test.

PR #85, opened 2019-04-05
fix component controller not finding gitsource schema

corinnekrych:
This PR fixes the issue in component controller:
- miss registration of the new gitsource schema
- fix make file for local dev

### DIFF
--- a/pkg/apis/devconsole/v1alpha1/zz_generated.deepcopy.go
+++ b/pkg/apis/devconsole/v1alpha1/zz_generated.deepcopy.go
@@ -13,7 +13,7 @@ func (in *Component) DeepCopyInto(out *Component) {
 	*out = *in
 	out.TypeMeta = in.TypeMeta
 	in.ObjectMeta.DeepCopyInto(&out.ObjectMeta)
-	in.Spec.DeepCopyInto(&out.Spec)
+	out.Spec = in.Spec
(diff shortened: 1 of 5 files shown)


tinakurian:
@corinnekrych thanks for fixing this

pkg/apis/devconsole/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -13,7 +13,7 @@ func (in *Component) DeepCopyInto(out *Component) {
 	*out = *in
 	out.TypeMeta = in.TypeMeta
 	in.ObjectMeta.DeepCopyInto(&out.ObjectMeta)
-	in.Spec.DeepCopyInto(&out.Spec)
```
akashshinde: @corinnekrych why do we need to change this ?

corinnekrych: I use latest version of operator-sdk v0.7.0
what about yours?

akashshinde: mine is v0.6.0, I was just wondering because there is a method `DeepCopyInto` on ComponentSpec struct. If we no longer using it anywhere shouldn't that be auto removed ? (not sure if we are using anywhere)

PR #87, opened 2019-04-06
Move Dockerfile.registry under the expected context

sbose78:
This makes it possible to setup continuous builds on Quay, to begin with.
Moreover, the Dockerfile is useful for generally building the registry irrespective of test scenarios.


baijum:
Please update `test-olm-integration` Makefile target in the `make/test.mk`.

sbose78:
Done https://github.com/redhat-developer/devconsole-operator/pull/87/commits/0c7c783420ff0752b56c237719e9ecdec19b06d9

PR #90, opened 2019-04-08
feat(ODC-464): add GitSourceAnalysis CRD

MatousJobanek:
Adds GitSourceAnalysis CRD. It will be a trigger of the build type detection and the result of the git repo analysis will be written into the status of this CRD.
Related PR adding API types to the devconsole-api repo: https://github.com/redhat-developer/devconsole-api/pull/8

### DIFF
--- a/deploy/crds/devconsole_v1alpha1_gitsourceanalysis_crd.yaml
+++ b/deploy/crds/devconsole_v1alpha1_gitsourceanalysis_crd.yaml
@@ -32,7 +32,7 @@ spec:
               minimum: 1
               properties:
                 name:
-                  description: Name is the name of the GitSource that contains
+                  description: Name is the name of the GitSource within the same namespace that contains
                     all necessary information of the git repo to by analyzed
                   type: string
                   minimum: 1
--- a/deploy/crds/devconsole_v1alpha1_gitsourceanalysis_crd.yaml
+++ b/deploy/crds/devconsole_v1alpha1_gitsourceanalysis_crd.yaml
@@ -1,0 +1,59 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: gitsourceanalyses.devconsole.openshift.io
+spec:
+  group: devconsole.openshift.io
+  names:
+    kind: GitSourceAnalysis
+    listKind: GitSourceAnalysisList
+    plural: gitsourceanalyses
+    singular: gitsourceanalysis
+  scope: Namespaced
+  validation:
+    # openAPIV3Schema is the schema for validating custom objects.
+    openAPIV3Schema:
+      properties:
+        apiVersion:
+          type: string
+        kind:
+          type: string
+        metadata:
+          type: object
+        spec:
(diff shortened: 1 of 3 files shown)


deploy/crds/devconsole_v1alpha1_gitsourceanalysis_crd.yaml
```diff
@@ -0,0 +1,83 @@
...
+        metadata:
+          type: object
+        spec:
+          properties:
+            gitSourceRef:
+              description: GitSourceRef refers to the GitSource to be analyzed
+              minimum: 1
+              properties:
+                name:
+                  description: Name is the name of the GitSource that contains
```
alexeykazakov:
```suggestion
                  description: Name is the name of the GitSource within the same namespace that contains
```

deploy/crds/devconsole_v1alpha1_gitsourceanalysis_crd.yaml
```diff
@@ -0,0 +1,83 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
alexeykazakov: We should really stop duplicating the same CRDs in deploy and manifest folders. We do need to store CRDs in manifests so let's try to reuse them everywhere instead of duplicating them in /deploy? **This is not for this PR to take care** though. Just saying :)

sbose78: yeah, the current way of storing the CRD yamls are so *tiring*.
Tests and everything needs to look at /deploy only for RBAC and manifests/<latest-version> for CRDs. 

Also, we need to handle the problem of rbac info staying in two places - /deploy and `clusterPermissions` in /manifests/<latest-version>

sbose78: tracked https://jira.coreos.com/browse/ODC-473

MatousJobanek:
> Also I think it's worth adding a few CR examples to https://github.com/redhat-developer/devconsole-operator/tree/master/examples

Thanks for your review @alexeykazakov. I added the example here a3e605e but didn't add the status part. Let me know if I should add it there too

sbose78:
@MatousJobanek , the `status` part could be included in a different example yaml I would suggest. This is because the example you added now is great for showing "how does a user express intent for analysis"

deploy/crds/devconsole_v1alpha1_gitsourceanalysis_crd.yaml
sbose78: Btw, 
FYI @baijum reminded me that there are sanity tests called as part of `make test` that verify that the CRDs are same.

alexeykazakov:
@sbose78 exactly.
@MatousJobanek I was thinking about a few examples with different status/build types. Failed. Analyzed with detected java/maven. Analyzed with golang... It would be helpful for the UI guys and other consumers to see different examples of what they can expect from the controller.

MatousJobanek:
ack @alexeykazakov I'll add them there

PR #91, opened 2019-04-08
feat(ODC-466): add status properties to GitSource CRD

MatousJobanek:
add status properties to GitSource CRD to be able to keep information about the state fo the CR and if the last connection attempt was successful or not
Related PR to the devconsole-api repo: https://github.com/redhat-developer/devconsole-api/pull/9


deploy/crds/devconsole_v1alpha1_gitsource_crd.yaml
```diff
@@ -58,18 +58,33 @@ spec:
...
+            connection:
+              description: Connection holds information whether the last attempt to reach the git source
+              was successful or not. Optional
+              properties:
+                state:
+                  description: State is the result of the attempt to reach a GitSource. Can be either Failed or OK
+                  type: string
+                  mimimum: 1
+                error:
+                  description: Error has the error message if the attempt to reach a GitSource failed
```
sbose78: @christianvogt
From a UI/consumption perspective,
1. if there's `error` and the `advanced -> private repo` options were hidden, the 
 `advanced -> private repo` should show up.

2.  if there's `error` and the `advanced -> private repo` options were NOT hidden, they should continue showing up.

christianvogt: Should we also capture a code to indicate whether the error is due to authentication, invalid url, unreachable, etc... ?
That way the UI can more clearly present the authentication section only when we detect it to be a private repository.

sbose78: We could consider including the `http` status code of the last attempt to reach the git url? Is that what you are suggesting @christianvogt ?
Or, are you looking for an enumerated list of error constants?

sbose78:
```
curl --verbose https://github.com/redhWRONGoper/devopsconsole-s?service=git-upload-pack
HTTP/1.1 401 Authorization Required
```

```
curl --verbose https://bitbucket.org/sbose78/tedxbangalore2017.git/info/refs?service=git-upload-pack
HTTP/2 401 
```

So, for both an incorrect URL and a private URL, the same response status is returned even thought the body had friendly textual information. 

I would say, let's not parse the text in the body to map to custom error codes since their could be way too many possibilities there. The client/UI could use the http response status code to hide/unhide parts?

christianvogt: If the status code is present, that would help.
401 we show the form to add auth.
404, or anything else, we simply show an error that the url is invalid.

MatousJobanek: @christianvogt unfortunately, GH responds with the same status code for both cases:
* when a user doesn't have sufficient permissions (eg. private repo) 
* when a user provided a wrong URL or branch. 

So in the case of GH, we cannot differentiate the failures

MatousJobanek:
> LGTM. But let's also add some examples to https://github.com/redhat-developer/devconsole-operator/tree/master/examples

I modified only the status part - should this be in the example as well? I though that the file contains example of a payload to be sent to OS, not example of the response...

deploy/crds/devconsole_v1alpha1_gitsource_crd.yaml
christianvogt: Ok. What about when the user enters the wrong domain altogether?

sbose78: @christianvogt I would be a bit careful with using the domain to identity anything because enterprise gitlab/github/bitbucket deployments have custom domains like source.ibm.com :)

sbose78:
@MatousJobanek , yeah. right. To @alexeykazakov 's point, we need the `status` examples because these 2 CRs are `status` heavy now. Therefore, we should include 'response' CRs as separate examples ( other than the payload CRs ).

deploy/crds/devconsole_v1alpha1_gitsource_crd.yaml
christianvogt: I was referring to status code when pinging a domain that doesn't exist at all. Not domain name inspection.

MatousJobanek:
sbose78 I've added some examples for those cases I currently know and I have implementation for. With the next controller functionality, I'll provide more examples.

As for your request regarding adding http status code of the response to the status - what would be this good for? What do you expect from this? AFAIK both GitHub and Gitlab respond the same response for both cases - when repo is either private or URL is wrong (only Bitbucket provide different responses). The only thing I can differentiate is:
* repo is not reachable (url is wrong or is private)
* repo is reachable but the branch does not exist

I'm not sure about the credentials validations - there will be most likely different response when you provide wrong credentials, but nothing more.

sbose78:
> As for your request regarding adding http status code of the response to the status - what would be this good for?

This could help the UI do a "if responseCode == 40x", show a generalized error message.

sbose78:
@MatousJobanek I'll merge this PR - https://github.com/redhat-developer/devconsole-operator/pull/118 looks good, but my concern is that the UI might want to use the returned value to show a more generic message.

While `unable to reach the repo` is helpful for drilling into details, the client might want to return the error text differently. :)

MatousJobanek:
> This could help the UI do a "if responseCode == 40x", show a generalized error message.

@sbose78 that could work if we got different response codes from the git serves for different types of failures. However, both GitHub and Gitlab respond with the same code for every failure. That's why I asked what that could be good for...
Instead of this, I'm proposing adding `error-type` or `failure-reason` to the status of the custom resource (along with or instead of the current `error`). This field would have one of the predefined set of possible values. UI would then match this `error-type` and react accordingly. I've tried to cover all the possible cases, responses and error-types in this document: https://docs.google.com/spreadsheets/d/1Qn2cioIXj8r3bcVddzISUZiJ2ODq6hWFg6TI_fIz34U/edit#gid=0
Please let me know if that make sense  (cc @christianvogt)

MatousJobanek:
This is something that I had in my mind from the very beginning, but I didn't want to introduce it in the first PR as I didn't know what kind of errors and levels of credentials validations I could handle. That's also the reason why I introduced just the long string for the time being.
For now,it seems that the list of errors should be final (at least for some time) so if @christianvogt agrees with it as well,then I'll send a PR with this new field.

christianvogt:
@MatousJobanek this looks good to me

PR #99, opened 2019-04-10
Validate manifests using operator-courier

baijum:
Resolves https://jira.coreos.com/browse/ODC-480

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,7 @@ RUN yum install epel-release -y \
     bc \
     kubectl \
     yamllint \
+    python36-virtualenv \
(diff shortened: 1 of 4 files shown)


baijum:
> Could we install operator-courier in the /Dockerfile and /openshift-ci/Dockerfile.tools ?

I see the `make test` is passing for this PR. Probably `python3` must be installed indirectly?

I am going to include `python36-virtualenv` into the `yum install` in both Dockerfile.

baijum:
@sbose78 I have included `python36-virtualenv` into the `yum install` in both Dockerfile.

PR #105, opened 2019-04-11
Refactor component

tinakurian:
Refactoring to remove component_types from devconsole-operator and use component_types in devconsole-api.


Gopkg.lock
```diff
@@ -362,7 +362,7 @@
   name = "github.com/redhat-developer/devconsole-api"
   packages = ["pkg/apis/devconsole/v1alpha1"]
   pruneopts = "NT"
-  revision = "337a67c0dc928d5a01a9f48d989d5ac74ef26e92"
+  revision = "a67d77335e36554ccf3fea971af2d0ff10af3161"
```
baijum: Instead of a commit hash, can we use the `master` branch here?

sbose78: It is usually unsafe to use master branches because the content of the master branch can vary between "code time" , "CI build time" and "merge to master" image generation time.

PR #106, opened 2019-04-11
Update.status

corinnekrych:
Fixes ODC-214
This PR takes care of:
- watching dc, bc, service, route as secondary resources, therefore if one of those get deleted afterwards, it gets respawned.
- update status of component to see if it is buildding or deploying.
- refactor GetgitSource to follow same pattern of get resources

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -83,6 +83,15 @@ required = [
   branch = "master"
   name = "github.com/wadey/gocovmerge"
 
+
 [[override]]
   name = "github.com/golang/protobuf"
   version = "1.3.1"
+
+[[constraint]]
+  name = "github.com/redhat-developer/devconsole-api"
+  packages = [
+    "pkg/apis",
+    "pkg/apis/devconsole/v1alpha1",
+  ]
+
--- a/pkg/controller/component/component_controller.go
+++ b/pkg/controller/component/component_controller.go
@@ -4,20 +4,17 @@ import (
 	"context"
 	e "errors"
 	"fmt"
-	"strconv"
-
 	v1 "github.com/openshift/api/apps/v1"
 	buildv1 "github.com/openshift/api/build/v1"
 	imagev1 "github.com/openshift/api/image/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	devconsoleapi "github.com/redhat-developer/devconsole-api/pkg/apis/devconsole/v1alpha1"
-
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
-
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -28,7 +25,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-var log = logf.Log.WithName("controller_component")
+var log = logf.Log
 
 // Add creates a new Component Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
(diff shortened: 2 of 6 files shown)


baijum:
@corinnekrych CRDs has been removed form `manifests` directory. During OLM e2e tests, those files are copied from `deploy`  directory.

My current thinking is, once we publicly release `0.1.0` release, the CRDs corresponding to that version can be placed there or moved to OperatorHub Red Hat Community operators.

corinnekrych:
@baijum yes i know I'll need to rebase and change my PR
it is still very much a WIP and since PRs are going fast in this repo i thought i'll let know the team what i'm working on...

sbose78:
> moved to OperatorHub Red Hat Community operators.

Did you mean the Github repo?
As far as I know, the community-operators github repo is a stop gap approach. We have to generate images - and 'link' them to the marketplace release. Have to figure out how.

sbose78:
We should also include the resources that were created by the component controller, just like:

https://github.com/operator-framework/operator-sdk-samples/blob/master/memcached-operator/pkg/controller/memcached/memcached_controller.go#L153

corinnekrych:
> We should also include the resources that were created by the component controller, just like:
> 
> https://github.com/operator-framework/operator-sdk-samples/blob/master/memcached-operator/pkg/controller/memcached/memcached_controller.go#L153

@sbose78 Why? I should have a list of the name of BuildConfig, DeploymentConfig, Service, Route?   what for ? we could just use label selector if we need those. Besides updating status required more code and I don't see the benefits. 
Watching the secondary resource so that if one get deleted it's up again with an indication of the status should be enough.

Gopkg.toml
```diff
@@ -81,3 +81,12 @@ required = [
 [[constraint]]
   branch = "master"
   name = "github.com/wadey/gocovmerge"
+
+[[constraint]]
+  branch = "update.status"
+  name = "github.com/redhat-developer/devconsole-api"
+  source = "https://github.com/corinnekrych/devconsole-api.git"
```
akashshinde: @corinnekrych shouldn't you use `github.com/redhat-developer/devconsole-git` here ?

corinnekrych: yes! this is temporary until this associated PR  get merged:https://github.com/redhat-developer/devconsole-api/pull/16 before merging i'll point to master

sbose78:
@corinnekrych Could you please add a detailed description of what specifically you are doing in this pull request?

pkg/controller/component/component_controller.go
```diff
@@ -77,9 +98,6 @@ type ReconcileComponent struct {
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileComponent) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
-	reqLogger.Info("Reconciling Component")
-
 	// Fetch the Component instance
 	instance := &devconsoleapi.Component{}
```
baijum: How about using `cr` as the variable name for Component instead of too generic `instance`?

corinnekrych: yes we could, let's do it in separate PR thou to avoid too many conflicts given we have some pending PR in the same code area.

sbose78:
> watching dc, bc, service, route as secondary resources, therefore if one of those get deleted afterwards, it gets respawned.
update status of component to see if it is buildding or deploying.

Approach is good for now!

Gopkg.toml
sbose78: Take it in https://github.com/redhat-developer/devconsole-api/pull/16

pkg/controller/component/component_controller.go
```diff
@@ -173,13 +303,13 @@ func (r *ReconcileComponent) CreateRoute(cr *devconsoleapi.Component) (*routev1.
 	foundRoute := &routev1.Route{}
 	err := r.client.Get(context.TODO(), types.NamespacedName{Name: route.Name, Namespace: route.Namespace}, foundRoute)
 	if err == nil {
-		log.Info("** Skip Creating CreateRoute: Already exist", "CreateRoute.Namespace", foundRoute.Namespace, "CreateRoute.Name", foundRoute.Name)
+		log.Info("** Skip Creating Route: Already exist", "Route.Namespace", foundRoute.Namespace, "Route.Name", foundRoute.Name)
 		return foundRoute, nil
 	}
 	if errors.IsNotFound(err) {
-		log.Info("** Creating a new CreateRoute", "CreateRoute.Namespace", route.Namespace, "CreateRoute.Name", route.Name)
+		log.Info("💡💡  Creating a new Route  💡💡", "Route.Namespace", route.Namespace, "Route.Name", route.Name)
```
sbose78: Could we removed the 💡 s? :)

pkg/controller/component/component_controller_test.go
```diff
@@ -81,6 +81,7 @@ func TestComponentController(t *testing.T) {
 			"password": []byte("password"),
 		},
 	}
+
```
tinakurian: can we remove this new line?

Gopkg.toml
```diff
@@ -83,6 +83,15 @@ required = [
...
 [[override]]
   name = "github.com/golang/protobuf"
   version = "1.3.1"
+
+[[constraint]]
+  name = "github.com/redhat-developer/devconsole-api"
+  packages = [
+    "pkg/apis",
+    "pkg/apis/devconsole/v1alpha1",
+  ]
```
MatousJobanek: @corinnekrych may I know what was the idea behind adding the `packages` key here? If I'm not mistaken, this is an invalid key for the `[[constraint]]` - packages are allowed only for `[[projects]]` in `Gopkg.lock` file

PR #110, opened 2019-04-12
Use `KUBECONFIG` instead of username & password

baijum:
The OpenShift 4 installer creates a Kubernetes configuration file for authentication.  The `oc` command and Operator SDK clients look for an environment variable named `KUBECONFIG` pointing to this configuration file. https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/

* Set OpenShift 4 as the default version as the test target

### DIFF
--- a/make/dev.mk
+++ b/make/dev.mk
@@ -12,8 +12,6 @@ DEVCONSOLE_OPERATOR_IMAGE?=quay.io/redhat-developers/devconsole-operator
 TIMESTAMP:=$(shell date +%s)
 TAG?=$(GIT_COMMIT_ID_SHORT)-$(TIMESTAMP)
 OPENSHIFT_VERSION?=4
-OC_LOGIN_USERNAME?=system:admin
-OC_LOGIN_PASSWORD?=admin
 
 .PHONY: create-resources
 create-resources:
--- a/make/dev.mk
+++ b/make/dev.mk
@@ -11,7 +11,7 @@ REGISTRY_URI=quay.io
 DEVCONSOLE_OPERATOR_IMAGE?=quay.io/redhat-developers/devconsole-operator
 TIMESTAMP:=$(shell date +%s)
 TAG?=$(GIT_COMMIT_ID_SHORT)-$(TIMESTAMP)
-OPENSHIFT_VERSION?=3
+OPENSHIFT_VERSION?=4
 OC_LOGIN_USERNAME?=system:admin
 OC_LOGIN_PASSWORD?=admin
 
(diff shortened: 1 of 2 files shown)


make/dev.mk
```diff
@@ -11,7 +11,7 @@ REGISTRY_URI=quay.io
 DEVCONSOLE_OPERATOR_IMAGE?=quay.io/redhat-developers/devconsole-operator
 TIMESTAMP:=$(shell date +%s)
 TAG?=$(GIT_COMMIT_ID_SHORT)-$(TIMESTAMP)
-OPENSHIFT_VERSION?=3
+OPENSHIFT_VERSION?=4
 OC_LOGIN_USERNAME?=system:admin
```
akashshinde: @baijum we can remove `OC_LOGIN_USERNAME` and `OC_LOGIN_PASSWORD` since we are no longer using it.

baijum: Good point :-)

baijum: Done d3e2600adb44d28f32b3010d70a9a7580df1260b

Avni-Sharma:
if I am logged in as `dev1 (or any other user group but admin) then the dev1 has access to only default namespace and no there namespace. Meaning no access to OLM namespace or operators namespace. But this test will need access to OLM namespace so we need to explicitly specify the OC_LOGIN as kubeadmin and the corresponding password. Even with the same KUBECONFIG file but as dev1 I cannot run the test. So the KUBECONFIG is not the only criteria. IMHO it will require to explicitly be a kubeadmin. I get this error on running as dev1
```
from server for: "STDIN": catalogsources.operators.coreos.com "my-catalog" is forbidden: User "user1" cannot get resource "catalogsources" in API group "operators.coreos.com" in the namespace "openshift-operator-lifecycle-manager"
make: *** [make/test.mk:74: test-olm-integration] Error 1
```
It might happen that in the Openshift environment that has been setup the user might be anything, so for safety purpose, we should ensure that the OC_LOGIN is set.

akashshinde:
@Avni-Sharma I am wondering would developer have access to catalogsource, Isn't it supposed to be accessible only to the cluster admins ?

sbose78:
Good point Avni!  Baiju is probably looking for the kubeconfig because the Openshift CI creates an ephemeral cluster and injects a kubeconfig into the test environment.

baijum:
The `oc` command also works with `KUBECONFIG` configuration. And it is used for cleanup. Whereas the Operator SDK framework expects `KUBECONFIG`. This makes `oc login` redundant.
https://github.com/operator-framework/operator-sdk/blob/master/doc/user-guide.md#2-run-locally-outside-the-cluster

Avni-Sharma:
kubeconfig would be provided as an environment variable then, I suppose..

baijum:
> kubeconfig would be provided as an environment variable then, I suppose..

Yes.

Avni-Sharma:
> > kubeconfig would be provided as an environment variable then, I suppose..
> 
> Yes.

cc @ldimaggi

baijum:
@akashshinde I need your approval to merge this PR.

PR #111, opened 2019-04-12
Remove operator-courier from test target

baijum:
The operator-courier has been moved to `lint` and CI has been enabled
for the same.


baijum:
The lint is enabled as a separate target in CI: https://github.com/openshift/release/pull/3438
Here is the `lint` target with `courier` and other "linters":
https://github.com/redhat-developer/devconsole-operator/blob/master/make/lint.mk#L11

PR #114, opened 2019-04-15
Adding labels and selectors

tinakurian:
Relates to: https://jira.coreos.com/browse/ODC-83
Require devconsole-api PR to me approved and merged first: https://github.com/redhat-developer/devconsole-api/pull/17/files in order to update the devconsole-api version in devconsole-operator.

Adding labels to custom resources.

To test: run component_controller_test.go tests.


pkg/resource/label.go
```diff
@@ -1,12 +1,44 @@
 package resource
 
-type NameGetter interface {
+// CRGetter is an interface which contains getter functions
+// to retrieve data from the custom resource.
+type CRGetter interface {
```
akashshinde: @tinakurian Where have you implemented `CRGetter` interface ?

akashshinde: @tinakurian Function names don't really convey that it's being used to get labels. How about having `GetLabelComponent`, `GetLabelInstance`, `GetLablePartOf` ? Reason I thought of this because when we implement this in any CR type it would be easier to understand this function's significance. 
What do you think ?

tinakurian: @akashshinde Thanks for the review! Updated it here: 3ed6948ce43c157b834d94c91544c0a7f8abe936 and here: https://github.com/redhat-developer/devconsole-api/pull/17

PR #117, opened 2019-04-15
Add GitSourceAnalysis controller

akashshinde:
This PR copies GitSourceAnalysis controller from https://github.com/redhat-developer/devconsole-git repo, Since we have decided to have all controller in one place.


pkg/controller/gitsourceanalysis/gitsourceanalysis_controller_test.go
```diff
@@ -0,0 +1,291 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/h2non/gock.v1"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	pathToTestDir  = "../../test"
```
akashshinde: @MatousJobanek Should I get `test` files from `vendor` here ? What do you you think ?

DhritiShikhar:
@akashshinde since we deicided to have all controller in one place, shouldn't there be a following PR to remove `GitSourceAnalysis` controller from  https://github.com/redhat-developer/devconsole-git ?

examples/devconsole_v1alpha1_gitsource_cr.yaml
```diff
@@ -9,6 +9,4 @@ spec:
   httpProxy: http://proxy.example.com
   httpsProxy: https://proxy.example.com
   noProxy: somedomain.com, otherdomain.com
-  secretRef:
```
DhritiShikhar: Why are these two lines removed?

pkg/controller/gitsourceanalysis/gitsourceanalysis_controller.go
```diff
@@ -0,0 +1,155 @@
...
+	if gsAnalysis.Status.Analyzed {
+		reqLogger.WithValues("git-source", gsAnalysis.Spec.GitSourceRef).
+			Info("Skipping GitSourceAnalysis as it was already analyzed")
+		return reconcile.Result{}, err
+	}
+
+	return analyze(reqLogger, r.client, gsAnalysis, request.Namespace)
+}
+
+func analyze(logger logr.Logger, client client.Client, gsAnalysis *v1alpha1.GitSourceAnalysis, namespace string) (reconcile.Result, error) {
```
DhritiShikhar: [SUGGESTION] It would be good to have a comment describing the function.

Same for following functions.

pkg/controller/gitsourceanalysis/gitsourceanalysis_controller.go
```diff
@@ -0,0 +1,155 @@
...
+	gitSource := &v1alpha1.GitSource{}
+	err := client.Get(context.TODO(), newNsdName(namespace, gsAnalysis.Spec.GitSourceRef.Name), gitSource)
+	if err != nil {
+		gsAnalysis.Status.Error = "Failed to fetch the input source"
+		logger.WithValues("git-source", gsAnalysis.Spec.GitSourceRef).
```
DhritiShikhar: Why do you use `logger` here and not [`controllerLogger`](https://github.com/redhat-developer/devconsole-operator/pull/117/files#diff-779d73a20f0e60d5e4b196bbda4a07a2R27) ?

pkg/controller/gitsourceanalysis/gitsourceanalysis_controller.go
```diff
@@ -0,0 +1,155 @@
...
+			Info("Skipping GitSourceAnalysis as it was already analyzed")
+		return reconcile.Result{}, err
+	}
+
+	return analyze(reqLogger, r.client, gsAnalysis, request.Namespace)
+}
+
+func analyze(logger logr.Logger, client client.Client, gsAnalysis *v1alpha1.GitSourceAnalysis, namespace string) (reconcile.Result, error) {
+	gitSource := &v1alpha1.GitSource{}
+	err := client.Get(context.TODO(), newNsdName(namespace, gsAnalysis.Spec.GitSourceRef.Name), gitSource)
```
DhritiShikhar: What is the need of this function `newNsdName`?
Can't we directly use `types.NamespacedName{...}`?

Refer: https://github.com/redhat-developer/devconsole-operator/blob/master/pkg/controller/component/component_controller.go#L174

pkg/controller/gitsourceanalysis/gitsourceanalysis_controller.go
```diff
@@ -0,0 +1,155 @@
...
+	if err != nil {
```
DhritiShikhar: If we reach this condition `err != nil`,
then there should be a return statement also `return nil, err`, right?

examples/devconsole_v1alpha1_gitsource_cr.yaml
corinnekrych: @akashshinde I think we should have 2 example files: one with secret named `devconsole_v1alpha1_gitsource_secret_cr.yaml` and this one (without secret). In my Pr I also removed it as it's not convenient when testing public repo.

akashshinde: @corinnekrych Yeah that's exactly what I was also thinking, I should also include `Secret` template for reference.

pkg/controller/gitsourceanalysis/gitsourceanalysis_controller.go
akashshinde: @DhritiShikhar I have just been refactoring this code, @MatousJobanek will be the right person to answer code related questions.

pkg/controller/gitsourceanalysis/gitsourceanalysis_controller.go
MatousJobanek: of course that we can, but this is shorter, I like it more and the function can be reused at more places ;-)

pkg/controller/gitsourceanalysis/gitsourceanalysis_controller.go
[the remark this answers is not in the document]
MatousJobanek: the idea was having it open for unit tests, but it's right that it is not used so far there. 
On the other hand, there is no big difference in the function declaration, so I think that we can keep it as it is.

pkg/controller/gitsourceanalysis/gitsourceanalysis_controller.go
MatousJobanek: why? the error msg is set to the GitSourceAnalysis CR that is then updated. If the error was returned, then the reconciler would be triggered again which is something that we don't want

pkg/controller/gitsourceanalysis/gitsourceanalysis_controller.go
MatousJobanek: because this one also contains information about the object name as well as the namespace name - look at the declaration of the logger that is passed to this function.

pkg/controller/gitsourceanalysis/gitsourceanalysis_controller.go
MatousJobanek: it's not an exported function, so I don't see it important to add the comment there

pkg/controller/gitsourceanalysis/gitsourceanalysis_controller.go
```diff
@@ -0,0 +1,155 @@
...
+	} else {
+		buildEnvStats, err := detector.DetectBuildEnvironments(logger, gitSource, gitSecretProvider)
+		if err != nil {
+			logger.Error(err, "Error detecting build types")
+		}
+		return buildEnvStats, err
+	}
+}
+
+func newNsdName(namespace, name string) types.NamespacedName {
```
sbose78: Could you please rename this to newNamespacedName(..) :D

sbose78:
@MatousJobanek & @DhritiShikhar  After your 👍 , let's merge this.

pkg/controller/gitsourceanalysis/gitsourceanalysis_controller.go
MatousJobanek: renamed ef3e10b

PR #118, opened 2019-04-15
feat(ODC-451): add GitSource controller that checks if repo is reachable and branch exists

MatousJobanek:
Adds GitSource controller that using a function from [devconsole-git](https://github.com/redhat-developer/devconsole-git/pull/23) checks if the repo is reachable or not. If it is not, then it means that either URL is wrong or the repo is private. In this case the operator sets:
*  `connection.state` to `failed` and `connection.error` to `unable to reach the repo`. 

If the repo is reachable then it checks if the provided branch exists. If it doesn't then it sets:
*  `connection.state` to `failed` and `connection.error` to `unable to find the branch`. 

If the repo is reachable and branch exists then it sets:
*  `connection.state` to `ok` and `connection.error` sets to empty string

this PR depends on:
https://github.com/redhat-developer/devconsole-git/pull/23 and https://github.com/redhat-developer/devconsole-operator/pull/91


pkg/controller/gitsource/gitsource_controller.go
```diff
@@ -0,0 +1,139 @@
...
+	return reconcile.Result{}, nil
+}
+
+func updateStatus(log *gslog.GitSourceLogger, gitSource *v1alpha1.GitSource) (isDirty bool) {
+
+	if gitSource.Status.Connection.State != "" {
+		return false
+	}
+	if gitSource.Status.State == "" {
+		gitSource.Status.State = v1alpha1.Initializing
```
sbose78: @MatousJobanek , this doesn't move to any other state, we should probably move this to `initialized` ?

MatousJobanek: The idea behind this is that the state should be either `ready` or `initialized` where `initialized` is the first step, then you can move it to `ready`. If the state is empty - the CR was just created - then we could assume that the CR is in `initialized` state - that is the reason why I set it here.

PR #119, opened 2019-04-16
Update version of github.com/golang/protobuf package to 1.3.1

pmacik:
Fixes https://jira.coreos.com/browse/ODC-565

Updating the `github.com/golang/protobuf` package to version `1.3.1` fixes that problem and the test runs again.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -83,6 +83,6 @@ required = [
   branch = "master"
   name = "github.com/wadey/gocovmerge"
 
-[[constraint]]
+[[override]]
   name = "github.com/golang/protobuf"
   version = "1.3.1"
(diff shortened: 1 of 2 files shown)


Gopkg.toml
```diff
@@ -81,3 +82,7 @@ required = [
 [[constraint]]
   branch = "master"
   name = "github.com/wadey/gocovmerge"
+
+[[constraint]]
```
dipak-pawar: I believe we are not using protobuf as direct dependency, we should use `[[override]]` here

pmacik: Thanks @dipak-pawar, that's a good point. I fixed that...

dipak-pawar: @pamcik, thanks, it'll be good if you can update `Gopkg.lock` to ensure a reproducible build?

PR #120, opened 2019-04-16
Fix duplicate target names

baijum:
These target names exist few lines above.

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -106,9 +106,9 @@ endif
 # e2e test in dev mode
 #-------------------------------------------------------------------------------
 
-.PHONY: e2e-cleanup
+.PHONY: e2e-cleanup-local
 ## Create a namespace used in e2e tests
-e2e-cleanup: get-test-namespace
+e2e-cleanup-local: get-test-namespace
 	$(Q)-oc login -u system:admin
 	$(Q)-oc delete -f ./deploy/crds/devconsole_v1alpha1_component_crd.yaml
 	$(Q)-oc delete -f ./deploy/service_account.yaml --namespace $(TEST_NAMESPACE)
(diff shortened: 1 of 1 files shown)


baijum:
I am planning to use the original targets to run tests with `--up-local` option.

make/test.mk
```diff
@@ -106,23 +106,23 @@ endif
...
-.PHONY: e2e-setup
+.PHONY: e2e-setup-local
 ## Create a namespace used in e2e tests
-e2e-setup: e2e-cleanup
+e2e-setup-local: e2e-cleanup-local
 	$(Q)-oc new-project $(TEST_NAMESPACE)
 
 .PHONY: build-image-local
-build-image-local: e2e-setup
+build-image-local: e2e-setup-local
```
dipak-pawar: @baijum, As we need `build-image-local` too frequently for develop and test phase, I would avoid cleaning up resources for each docker build, otherwise to test some change I need to wait to clean up all openshift resources which adds more than few seconds.

make/test.mk
```diff
@@ -106,23 +106,23 @@ endif
 # e2e test in dev mode
 #-------------------------------------------------------------------------------
 
-.PHONY: e2e-cleanup
+.PHONY: e2e-cleanup-local
```
DhritiShikhar: [e2e-cleanup-local](https://github.com/redhat-developer/devconsole-operator/pull/120/files#diff-57dae1b5b313e109d7ebdfd77a500c5cR109) is similar to [e2e-cleanup](https://github.com/redhat-developer/devconsole-operator/pull/120/files#diff-57dae1b5b313e109d7ebdfd77a500c5cR52) . Can't we remove one of them?

make/test.mk
baijum: @dipak-pawar We can fix that in a separate PR.  It might require more discussion. For now, let me clear this name collision.

Avni-Sharma: but don't we need a clean slate before any test?

make/test.mk
baijum: @DhritiShikhar I will address this in a separate PR.

DhritiShikhar: @baijum okay but what is the need of this change then?

baijum: @DhritiShikhar To avoid ambiguous Makefile target names.  Same names are repeated. Your question is more the content within the target. That we can address separately.

PR #121, opened 2019-04-16
Make test-e2e to run test with local operator

baijum:
Use `--up-local` option to run the test

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -44,11 +44,11 @@ endif
 
 .PHONY: e2e-setup
 e2e-setup: e2e-cleanup 
-	$(Q)-oc new-project $(TEST_NAMESPACE)
+	$(Q)oc new-project $(TEST_NAMESPACE)
 
 .PHONY: e2e-cleanup
 e2e-cleanup: get-test-namespace
-	$(Q)-oc delete project $(TEST_NAMESPACE)  --wait
+	$(Q)-oc delete project $(TEST_NAMESPACE) --timeout=10s --wait
 
 .PHONY: test-olm-integration
 ## Runs the OLM integration tests without coverage
--- a/make/test.mk
+++ b/make/test.mk
@@ -36,6 +36,9 @@ get-test-namespace: ./out/test-namespace
 test-e2e: e2e-setup
 	$(info Running E2E test: $@)
 	$(eval DEPLOYED_NAMESPACE := $(TEST_NAMESPACE))
+ifeq ($(OPENSHIFT_VERSION),3)
+	$(Q)oc login -u system:admin
+endif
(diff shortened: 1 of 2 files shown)


make/test.mk
```diff
@@ -32,33 +32,20 @@ get-test-namespace: ./out/test-namespace
...
 	$(info Running E2E test: $@)
-	$(Q)go test ./test/e2e/... \
-		-parallel=1 \
-		${V_FLAG} \
-		-root=$(PWD) \
-		-kubeconfig=$(HOME)/.kube/config \
-		-globalMan ./deploy/test/global-manifests.yaml \
-		-namespacedMan ./deploy/test/namespace-manifests.yaml \
-		-singleNamespace
+	$(eval DEPLOYED_NAMESPACE := $(TEST_NAMESPACE))
```
DhritiShikhar: @baijum what is the need to store the `TEST_NAMESPACE` value as `DEPLOYED_NAMESPACE` and then use it in the next line?

make/test.mk
```diff
@@ -32,33 +32,23 @@ get-test-namespace: ./out/test-namespace
...
-## TODO: TBD
 e2e-cleanup: get-test-namespace
-	$(Q)-oc login -u system:admin
-	$(Q)-oc delete -f ./deploy/crds/devconsole_v1alpha1_component_crd.yaml
-	$(Q)-oc delete -f ./deploy/crds/devconsole_v1alpha1_gitsource_crd.yaml
-	$(Q)-oc delete -f ./deploy/service_account.yaml --namespace $(TEST_NAMESPACE)
-	$(Q)-oc delete -f ./deploy/role.yaml --namespace $(TEST_NAMESPACE)
-	$(Q)-oc delete -f ./deploy/test/role_binding_test.yaml --namespace $(TEST_NAMESPACE)
-	$(Q)-oc delete -f ./deploy/test/operator_test.yaml --namespace $(TEST_NAMESPACE)
+	$(Q)-oc delete project $(TEST_NAMESPACE)  --wait
```
DhritiShikhar: Is there a need to add `--timeout` here as well?

make/test.mk
baijum: @DhritiShikhar That `DEPLOYED_NAMESPACE` environment variable is later used in the test. The test was already written before. If you see my change in `test/e2e/component_test.go` in this PR, you can see the location where it is used.  The environment variable is used to identify the namespaces. Later inside the test, a wait is added for the operator deployment to take place.

make/test.mk
baijum: You mean the `--wait` argument? That is required to ensure the project (namespace) has been deleted.

make/test.mk
DhritiShikhar: addressed here 
efa601c

make/test.mk
DhritiShikhar: addressed here 
4aaa60c

PR #122, opened 2019-04-16
adding annotations

tinakurian:
Relates to: https://jira.coreos.com/browse/ODC-83
Require devconsole-api PR to me approved and merged first: https://github.com/redhat-developer/devconsole-api/pull/19

Adding annotations to custom resources.

To test: run component_controller_test.go tests.


pkg/controller/component/component_resources.go
```diff
@@ -121,11 +126,13 @@ func (r *ReconcileComponent) newBuildConfig(cr *devconsoleapi.Component, builder
 
 func newDeploymentConfig(cr *devconsoleapi.Component, output *imagev1.ImageStream) *v1.DeploymentConfig {
```
DhritiShikhar: `Annotations` can also be added to `ObjectMeta`:

```
ObjectMeta: metav1.ObjectMeta{                                                                                                                            
                           Name:      cr.Name,                                                                                                                               
                           Namespace: cr.Namespace,                                                                                                                          
                           Labels:    labels,      
                           Annotations: annotations,                                                                                                      
             },   
```



at line 147 (https://github.com/redhat-developer/devconsole-operator/pull/122/files#diff-3240f219b66b3cb310b8835fe5cf7e42R144)

tinakurian:
> /retest

Thanks @baijum! This won't pass until i can update the revision for devconsole-api. Will only be able to do that after this PR is merged: https://github.com/redhat-developer/devconsole-api/pull/19. However, the ci for devconsole-api is currently not working :-) so will need to fix that first.

pkg/controller/component/component_resources.go
tinakurian: Good catch @DhritiShikhar will update!

tinakurian: fixed here: e3c82fdee4dab0d861e8ee8e974bea9795a64c4d

tinakurian:
**Results from running component_controller_test.go:**

image

PR #130, opened 2019-04-17
Exposed ports

corinnekrych:
Fixes odc-515
This PR makes sure the default port is not hard coded to 8080 but is extracted from S2I builder image.
- If the port is provided in Component's spec port, this port is used instead of the S2I exposed port.
- if no port is provided in Component's spec port, port is extracted from S2I builder image.

### DIFF
--- a/pkg/controller/component/utils.go
+++ b/pkg/controller/component/utils.go
@@ -1,0 +1,78 @@
+package component
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/openshift/api/image/docker10"
+	imagev1 "github.com/openshift/api/image/v1"
+	corev1 "k8s.io/api/core/v1"
+	"strconv"
+	"strings"
+)
+
+// getExposedPortsFromISI parse ImageStreamImage definition and return all exposed ports in form of ContainerPorts structs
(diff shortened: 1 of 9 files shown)


pkg/controller/component/utils.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/openshift/api/image/docker10"
+	imagev1 "github.com/openshift/api/image/v1"
+	corev1 "k8s.io/api/core/v1"
+	"strconv"
+	"strings"
+)
+
+// getExposedPortsFromISI parse ImageStreamImage definition and return all exposed ports in form of ContainerPorts structs
+// borrowed from https://github.com/openshift/odo/blob/2b98a06e811a60a08c9a6545c9698896d91e8f0b/pkg/occlient/occlient.go#L541
+func getExposedPortsFromISI(image *imagev1.ImageStreamImage) ([]corev1.ContainerPort, error) {
```
sbose78: Could you use `getExposedPortsFromImageStreamImage` ?

corinnekrych: Done in 8a7d34a

PR #131, opened 2019-04-19
Add Travis CI configuration for e2e test

baijum:
The test runs on a Kubernetes cluster using Minikube


baijum:
This is an experiment. If it doesn't work we can disable it. This check will not be a blocker for any PR for now.

sbose78:
Thanks, which Travis instance picks up this? Would you be adding a webhook?

baijum:
> Thanks, which Travis instance picks up this? Would you be adding a webhook?

We should configure Travis Webhook for this repo. I can do that.

For testing this configuration, I configured it with my fork of this repo. See the results here:
https://travis-ci.org/baijum/devconsole-operator

PR #132, opened 2019-04-19
replace general instance by component cp

corinnekrych:
@baijum as discussed in https://github.com/redhat-developer/devconsole-operator/pull/106#discussion_r276063091

### DIFF
--- a/pkg/controller/component/component_controller.go
+++ b/pkg/controller/component/component_controller.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	e "errors"
 	"fmt"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "github.com/openshift/api/apps/v1"
 	buildv1 "github.com/openshift/api/build/v1"
 	imagev1 "github.com/openshift/api/image/v1"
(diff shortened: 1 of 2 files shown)


pkg/controller/component/component_controller.go
```diff
@@ -104,8 +104,8 @@ type ReconcileComponent struct {
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileComponent) Reconcile(request reconcile.Request) (reconcile.Result, error) {
 	// Fetch the Component instance
-	instance := &devconsoleapi.Component{}
-	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
+	cp := &devconsoleapi.Component{}
```
sbose78: What are your thoughts on using `component` instead of `cp` 
- doesn't collide with any other identifier. 
- `cp` has a different semantic meaning as well. 'Copy'

corinnekrych: I like `cp` as it is the short name of our component resource.

An interesting discussion on naming with idiomatic Go
>  "use short variable names over descriptive ones"

from http://wordaligned.org/articles/go-for-short-variable-names

PR #133, opened 2019-04-19
Simplify.and.update.makefile

corinnekrych:
@akashshinde this PR does:
- add deployment of GitAnalysis CR
- remove duplicate targets in makefile

PR #137, opened 2019-04-23
Register gitsourceanalysis to deploy-test target

pratikjagrut:
`make deploy-test` should deploy component, gitsource and gitsourceanalysis cr. But I saw gitsourceanalysis was missing so this PR is to add it to deploy-test target.

Steps to test:
`make clean`
`make build`
`make local`
after seeing some msg like: `{"level":"info","ts":1556020410.5623667,"msg":"🎉🎉  Component myapp has been successfully created!  🎉🎉 "}` 
in new terminal window run `make deploy-clean` and then `make deploy-test`
and final step is to run `oc get gs, gsa` 
you should see something like this
```
NAME                                                  STATUS
gitsource.devconsole.openshift.io/example-gitsource   

NAME                                                                   STATUS
gitsourceanalysis.devconsole.openshift.io/example-gitsource-analysis   true

```

### DIFF
--- a/make/deploy.mk
+++ b/make/deploy.mk
@@ -45,6 +45,7 @@ deploy-test:
 	$(Q)-oc new-project $(LOCAL_TEST_NAMESPACE)
 	$(Q)-oc apply -f examples/devconsole_v1alpha1_gitsource_cr.yaml
 	$(Q)-oc apply -f examples/devconsole_v1alpha1_component_cr.yaml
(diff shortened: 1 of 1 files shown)


pratikjagrut:
Duplicate of [PR133](https://github.com/redhat-developer/devconsole-operator/pull/133/files#diff-fd8f90289cf39bfe347fc71f872e5861R57)

PR #138, opened 2019-04-23
 feat(ODC-64): modify GS controller to validate credentials,repository and branch

MatousJobanek:
when the secret is specified in the `GitSource` then the controller validates:
* credentials
* repository
* branch

If any of the validation fails, then it sets the appropriate reason as well as the error message.
For more information see https://github.com/redhat-developer/devconsole-git/pull/26

#### Blockers:
* This PR is built on top of https://github.com/redhat-developer/devconsole-operator/pull/118, so the only commit to be reviewed is: 4a5045f 
* This PR depends on https://github.com/redhat-developer/devconsole-git/pull/26


MatousJobanek:
Closing and moving here https://github.com/redhat-developer/devconsole-git/pull/28 as the controllers were moved to `devconsole-git` repo.

PR #142, opened 2019-04-24
Updating devconsole api revision

tinakurian:
Updating devconsole-pai revision. 

Latest revision uses the buildtype as the default value for app.kubernetes.io/name if not present


tinakurian:
> We should only update Gopkg.toml like this.
> 
> ```
> [[constraint]]
>   name = "github.com/redhat-developer/devconsole-api"
>   ...
>   
>   revision = "e9de16b0f6bbc760546eb19e3cec6e043855f8fa"
> ```

i don't think we have ever had a revision in the .toml file :S adding it..

> We should only update Gopkg.toml like this.
> 
> ```
> [[constraint]]
>   name = "github.com/redhat-developer/devconsole-api"
>   ...
>   
>   revision = "e9de16b0f6bbc760546eb19e3cec6e043855f8fa"
> ```

Wasn't aware that we added it to the .toml a couple days ago: https://github.com/redhat-developer/devconsole-operator/commit/9d67273452d64580e9ff25bf175e615a2d806526#diff-836546cc53507f6b2d581088903b1785

Updated with revision in .toml file

PR #144, opened 2019-04-25
Update.version.runtime controller

corinnekrych:
This PR is about:
- updating Status with subresource so that Update only updates status and not the whole component
- update runtime-controller to v0.2.0 (i notice it makes update subresource more stable)
- use instance label to select dc, bc


Gopkg.toml
```diff
@@ -91,19 +85,14 @@ required = [
...
   name = "github.com/golang/protobuf"
   version = "1.3.1"
 
 [[constraint]]
   name = "github.com/redhat-developer/devconsole-api"
-  packages = [
-    "pkg/apis",
-    "pkg/apis/devconsole/v1alpha1",
-  ]
+  source = "github.com/corinnekrych/devconsole-api"
```
tinakurian: Should probably point to redhat-developer/devconsole-api instead of corinnekrych/devconsole-api?

corinnekrych: yes. this is temporary. will remove it once https://github.com/redhat-developer/devconsole-api/pull/23 is merged

PR #148, opened 2019-04-26
Add binaries to build registry into root image

sbose78:
- added a new image to build registry ( if this works, we can unify both )
- plan is to use the base image built in openshift-ci/ while building the registry

### DIFF
--- a/openshift-ci/Dockerfile.registry.build
+++ b/openshift-ci/Dockerfile.registry.build
@@ -1,0 +1,15 @@
+FROM quay.io/openshift/origin-operator-registry:latest
+FROM registry.access.redhat.com/ubi7-dev-preview/ubi-minimal:latest
+
+
+ARG version=0.1.0
+
+COPY manifests manifests
+COPY deploy/crds/*.yaml manifests/devconsole/${version}/
+
+USER root
+RUN sed -e "s,REPLACE_IMAGE,registry.svc.ci.openshift.org/$(OPENSHIFT_BUILD_NAMESPACE)/stable:devconsole-operator," -i manifests/devconsole/${version}/devconsole-operator.v${version}.clusterserviceversion.yaml
+USER 1001
+
+RUN initializer
+CMD ["registry-server", "--termination-log=log.txt"]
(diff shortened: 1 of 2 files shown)


openshift-ci/Dockerfile.registry.build
```diff
@@ -0,0 +1,15 @@
+FROM quay.io/openshift/origin-operator-registry:latest
```
sbose78: This would be replaced by the `FROM root` in openshift-ci

openshift-ci/Dockerfile.tools
```diff
@@ -54,6 +54,11 @@ RUN mkdir -p $GOPATH/src/github.com/operator-framework \
     && make dep \
     && make install
 
+COPY --from=quay.io/openshift/origin-operator-registry:latest /bin/initializer /bin/initializer
```
sbose78: Because `FROM quay.io/openshift/origin-operator-registry:latest` would be replaced, we need these binaries.

openshift-ci/Dockerfile.registry.build
```diff
@@ -0,0 +1,15 @@
...
+FROM registry.access.redhat.com/ubi7-dev-preview/ubi-minimal:latest
+
+
+ARG version=0.1.0
+
+COPY manifests manifests
+COPY deploy/crds/*.yaml manifests/devconsole/${version}/
+
+USER root
+RUN sed -e "s,REPLACE_IMAGE,registry.svc.ci.openshift.org/$(OPENSHIFT_BUILD_NAMESPACE)/stable:devconsole-operator," -i manifests/devconsole/${version}/devconsole-operator.v${version}.clusterserviceversion.yaml
```
sbose78: Since we would be doing a ` FROM root`, the `$(OPENSHIFT_BUILD_NAMESPACE)` would be available..

sbose78:
Merging because 
```
2019/04/26 17:00:58 Building root
2019/04/26 17:05:33 Build root succeeded after 4m32s
2019/04/26 17:05:35 Building src
2019/04/26 17:05:53 Build src succeeded after 16s
```

PR #149, opened 2019-04-26
Update env reference in registry builder

sbose78:


openshift-ci/Dockerfile.registry.build
```diff
@@ -8,8 +8,9 @@ COPY manifests manifests
 COPY deploy/crds/*.yaml manifests/devconsole/${version}/
 
 USER root
-RUN sed -e "s,REPLACE_IMAGE,registry.svc.ci.openshift.org/$(OPENSHIFT_BUILD_NAMESPACE)/stable:devconsole-operator," -i manifests/devconsole/${version}/devconsole-operator.v${version}.clusterserviceversion.yaml
+RUN ls /bin && sed -e "s,REPLACE_IMAGE,registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:devconsole-operator," -i manifests/devconsole/${version}/devconsole-operator.v${version}.clusterserviceversion.yaml
+RUN /bin/initializer
```
baijum: Should we add `USER 1001` before this?

sbose78: That's what I had, trying to see if that's what caused the issue.. hence moved it up..

baijum: ok

sbose78:
I'll see if I can get some answers on the CI log before we merge this :)

baijum:
> I'll see if I can get some answers on the CI log before we merge this :)

ok!

sbose78:
I think instead of
```
FROM quay.io/openshift/origin-operator-registry:latest
FROM registry.access.redhat.com/ubi7-dev-preview/ubi-minimal:latest
```
( which is horribly wrong anyway :D )
i should use 
```
FROM quay.io/openshift/origin-operator-registry:latest
```
So that openshift can override the first one and use the pipeline image. Probably, now the pipeline image isn't getting used. Rather `FROM registry.access.redhat.com/ubi7-dev-preview/ubi-minimal:latest` gets used which doesn't have the stuff inside `/bin`

baijum:
> I think instead of
> 
> ```
> FROM quay.io/openshift/origin-operator-registry:latest
> FROM registry.access.redhat.com/ubi7-dev-preview/ubi-minimal:latest
> ```
> ( which is horribly wrong anyway :D )
> i should use
> 
> ```
> FROM quay.io/openshift/origin-operator-registry:latest
> ```
> So that openshift can override the first one and use the pipeline image. Probably, now the pipeline image isn't getting used. Rather `FROM registry.access.redhat.com/ubi7-dev-preview/ubi-minimal:latest` gets used which doesn't have the stuff inside `/bin`

That makes sense!

openshift-ci/Dockerfile.registry.build
baijum: Now, Do you want to move this down again?

sbose78:
Taking this in, because the image build of operator-registry isn't in master of openshift/release.

PR #151, opened 2019-04-26
Allow the INITIALIZER Abinary to be found from PATH

sbose78:


sbose78:
Merging because this doesn't come in the path of any CI test

PR #171, opened 2019-04-30
Increase timeout to 30 minute

baijum:


baijum:
Lint is failing for many jobs. The e2e tests longer time, so I don't see any problem for increasing to 30 minutes. Later we can fine tune used lints and reduce time.

PR #174, opened 2019-05-02
Create make target to push operator to the quay app registry for tests.

pmacik:
This PR:
* Adds a make target (`push-operator-app-registry`) to push freshly built operator to the application registry in quay.io which can be used by `OperatorSource`
* Includes the `tmp` directory created by the above target in the `clean` target
* Adds a make target (`get-operator-version`) that detects the operator version from `./manifests/devconsole/devconsole.package.yaml` file and sets the `DEVCONSOLE_OPERATOR_VERSION` variable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ build: ./out/operator
 clean:
 	$(Q)-rm -rf ${V_FLAG} ./out
 	$(Q)-rm -rf ${V_FLAG} ./vendor
+	$(Q)-rm -rf ${V_FLAG} ./tmp
 	$(Q)go clean ${X_FLAG} ./...
 
 ./vendor: Gopkg.toml Gopkg.lock
--- a/make/test.mk
+++ b/make/test.mk
@@ -8,7 +8,6 @@ include ./make/out.mk
 # Quay App Registry
 DEVCONSOLE_APPR_NAMESPACE ?= odcqe
 DEVCONSOLE_APPR_REPOSITORY ?= devconsole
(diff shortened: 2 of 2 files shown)


pmacik:
> The changes look find - but make push-operator-app-registry results in a docker image being created for the operator.

@ldimaggi Yes, that is because the `push-operator-image` target is called before (as a dependency). The idea is to build fresh local operator image, tag it with the generated unique name - including the `$(TAG)` string - push it to the quay.io image repository and then reference this unique image in the operator manifest that is then used to package the operator application and push it to the app registry with that unique name. Then, the unique operator application can be referenced in the `OperatorSource`.

This approach makes it possible to test the current state of the code of the operator and to test it locally or automated in a "per-PR" manner.

ldimaggi:
The "REPLACE_IMAGE" placeholder here must also be updated before the operator is pushed to quay.io:

https://github.com/redhat-developer/devconsole-operator/blob/master/manifests/devconsole/0.1.0/devconsole-operator.v0.1.0.clusterserviceversion.yaml#L82

pmacik:
> The "REPLACE_IMAGE" placeholder here must also be updated before the operator is pushed to quay.io:

@ldimaggi It is updated - the placeholder is already being replaced. That is what [this line](https://github.com/redhat-developer/devconsole-operator/pull/174/files#diff-57dae1b5b313e109d7ebdfd77a500c5cR91) is for :)

pmacik:
> Any luck with figuring out how to add secrets to openshift ci?

Haven't tried, yet.

PR #175, opened 2019-05-02
Use the base image available in the OpenShift CI

baijum:
This confuration should work with any operator registry location.
The location is provided as an arument with default value pointing to
OpenShift CI.

### DIFF
--- a/openshift-ci/Dockerfile.registry.build
+++ b/openshift-ci/Dockerfile.registry.build
@@ -1,12 +1,11 @@
 FROM quay.io/openshift/origin-operator-registry:latest
 
-ARG image=registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:devconsole-operator
 ARG version=0.1.0
 
 COPY manifests manifests
 COPY deploy/crds/*.yaml manifests/devconsole/${version}/
 
-RUN sed -e "s,REPLACE_IMAGE,${image}," -i manifests/devconsole/${version}/devconsole-operator.v${version}.clusterserviceversion.yaml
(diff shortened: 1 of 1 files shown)


openshift-ci/Dockerfile.registry.build
```diff
@@ -1,24 +1,14 @@
 FROM quay.io/openshift/origin-operator-registry:latest
 
+ARG image=registry.svc.ci.openshift.org/${OPENSHIFT_BUILD_NAMESPACE}/stable:devconsole-operator
```
sbose78: Well, only the following are receptive towards environment variables. 
```
ADD
COPY
ENV
EXPOSE
FROM
LABEL
STOPSIGNAL
USER
VOLUME
WORKDIR
```

sbose78: https://docs.docker.com/engine/reference/builder/#environment-replacement

openshift-ci/Dockerfile.registry.build
```diff
@@ -1,24 +1,14 @@
...
 ARG version=0.1.0
 
 COPY manifests manifests
 COPY deploy/crds/*.yaml manifests/devconsole/${version}/
-RUN curl -L -s https://github.com/devconsole-operator-test-deps/operator-registry-binaries/releases/download/v0.1.0/appregistry-server -o /bin/appregistry-server \
```
sbose78: Let's keep only these changes and have this PR merged?

baijum: Done. 697d40640b29644386afa4d44424d661f819603e

PR #176, opened 2019-05-03
Set registry log

baijum:


baijum:
I want to merge this to test: https://github.com/openshift/release/pull/3634

PR #189, opened 2019-05-09
Remove unnecessary container ready badge.

akashshinde:


baijum:
@sbose78 The registry image is created for automated test purpose. Do we need that badge here?

sbose78:
Today, we use this registry image for installing the operator on the cluster using the `CatalogSource`

PR #190, opened 2019-05-09
Create separate Makefile target to update CSV

baijum:


baijum:
This is required as the Makefile macro expansion (`eval`) happens before running other commands.

PR #194, opened 2019-05-10
Print DevConsole operator version

baijum:


cmd/manager/main.go
```diff
@@ -32,6 +33,7 @@ func printVersion() {
 	log.Info(fmt.Sprintf("Go Version: %s", runtime.Version()))
 	log.Info(fmt.Sprintf("Go OS/Arch: %s/%s", runtime.GOOS, runtime.GOARCH))
 	log.Info(fmt.Sprintf("operator-sdk Version: %v", sdkVersion.Version))
+	log.Info(fmt.Sprintf("DevCosole Operator Version: %v", version.Version))
```
sbose78: Could we also print the commit id ?

baijum: I have added a ticket for this: https://jira.coreos.com/browse/ODC-798
Also asked a question related to this task here:
https://coreos.slack.com/archives/C3VS0LV41/p1557463827398600

PR #195, opened 2019-05-10
WIP : Deploy the App Topology Rest Service

sbose78:
In this pull request, 
- a Deployment is created in the `openshift-operators` namespace such that 
- the corresponding `service` would be reachable by `devconsole-app.openshift-operators.svc:8080`

Calls to this `service` would be proxied from the console backend [1] .

- [ ] Verify that console backend is able to proxy to this
- [ ] Create Service
- [ ] Watch and reconcile
- [ ] Add tests
- [ ] Parameterize the image name that is being deployed

https://github.com/openshift/console/blob/master/server/server.go#L34

### DIFF
--- a/pkg/service/topology/topology.go
+++ b/pkg/service/topology/topology.go
@@ -1,0 +1,170 @@
+package topology
+
+import (
+	"fmt"
+
+	v1 "github.com/openshift/api/apps/v1"
+	deploymentconfig "github.com/openshift/client-go/apps/clientset/versioned/typed/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


pkg/service/topology/topology.go
```diff
@@ -0,0 +1,168 @@
...
+		}}
+	}
+	return &v1.DeploymentConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ServiceName,
+			Namespace: ServicesNamespace,
+			Labels:    labels,
+			//	Annotations: annotations,
+		},
+		Spec: v1.DeploymentConfigSpec{
```
baijum: Why not Deployment?

sbose78: Right. We could probably make it a `Deployment` since this is not going to use `ImageStream`s

PR #204, opened 2019-05-13
Add controller to deploy topology service

akashshinde:
This PR borrows the work from https://github.com/redhat-developer/devconsole-operator/pull/195

A Deployment is created in the `openshift-operators` namespace such that
the corresponding service would be reachable at devconsole-app.openshift-operators.svc:8080
Calls to this service would be proxied from the console backend [1] .

To Do 
  - [ ] Verify that console backend is able to proxy to this
  - [ ] Create Deployment 
  - [ ] Create Service
  - [ ] Watch and reconcile
  - [ ] Parameterize the image name that is being deployed
  - [ ] Add tests  
       - [ ] Add unit tests
       - [ ] Add e2e tests


tinakurian:
@akashshinde can we please add some tests for these changes?

sbose78:
> A Deployment is created in the `openshift-operators`

@Avni-Sharma Can we deploy it in a different namespace?

akashshinde:
@sbose78 How do you suggest to parameterize docker image path of topology service ?

pkg/controller/service/topology/topology_controller.go
```diff
@@ -0,0 +1,246 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      ServiceName,
+					Namespace: ServicesNamespace,
+					Labels:    labels,
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{
+						{
+							Name:  ServiceName,
+							Image: "quay.io/redhat-developer/app-service:latest", // TODO(Akash): parameterize this
```
akashshinde: [Note]: This path will be parameterized in another PR

baijum: This is a big technical debt!

baijum:
> > A Deployment is created in the `openshift-operators`
> 
> @Avni-Sharma Can we deploy it in a different namespace?

Let's create a new dedicated namespace for our service(s).

pkg/controller/service/topology/topology_controller.go
baijum: How about creating a `DevConsoleConfig` CRD and keep the docker image as an attribute there?
See this Console CRD for example https://github.com/openshift/console-operator/blob/master/manifests/00-crd-operator-config.yaml

PR #211, opened 2019-05-16
Add a note about YUM deps and pull request order

baijum:


openshift-ci/README.md
```diff
@@ -79,6 +79,12 @@ canonical_go_repository: github.com/redhat-developer/devconsole-operator
 The above line ensures the cloned source code goes into the specified
 path: `$GOPATH/src/<canonical_go_repository>`.
 
+Note: If your pull request depends on any package installed through
+`dnf install` inside the `root` image, those changes should be sent
```
pmacik: @baijum There is `yum` instead of `dnf` in the base images such as `centos:7`.

baijum: Fixed here: 7693021b980364767760c186200a9c2700ec6f98

PR #214, opened 2019-05-20
replace the hard sleep

Avni-Sharma:


ldimaggi:
Note that additional checking will be implemented here:  https://jira.coreos.com/browse/ODC-830

akashshinde:
@Avni-Sharma Can you please add description to PR.

PR #206, opened 2019-05-14
[WIP] Create UI test to verify Developer perspective

pmacik:
This PR adds a new UI test that:
* Installs latest OS console
* Installs Dev Console
* Adds a non-admin user
* Verifies that Developer Perspective appears in UI for the user after the Dev Console is installed:
  * Login the user in UI
  * Check for the perspective switcher button in UI
  * Verify the `Developer` perspective is present

PR #215, opened 2019-05-20
Install htpasswd to the tools images for CI

pmacik:
This PR installs `htpasswd` tool (via `httpd-tools` package) to the tools image for CI.


pmacik:
> This looks fine - but how/where is this used?

@ldimaggi It is used in [here](https://github.com/redhat-developer/devconsole-operator/blob/master/hack/install_devconsole/create_user.sh#L13), and the script is used by the #206 test that will run on CI

